PR #3, opened 2026-04-12
fix: enforce authz on exposed endpoints and allow DELETE method in CORS

tteon:
## Feature
This PR adds missing authorization checks to multiple sensitive endpoints and fixes a CORS misconfiguration.

## Why
These endpoints were previously accessible without enforcing runtime policies. Specifically:
- `list_databases`, `list_graphs`, `list_agents` exposed global registry information.
- `platform_chat_session_get` and `platform_chat_session_reset` lacked policy checks entirely, presenting a risk to privacy and application integrity.
Additionally, the `platform_chat_session_reset` endpoint relies on the HTTP `DELETE` method which was not permitted by the `CORSMiddleware`, causing browser requests to fail.

## Design
- `workspace_id: str = Query(default="default", ...)` was added to the signature of these unprotected endpoints to preserve backward compatibility (similar to other endpoints).
- `require_runtime_permission()` calls were added to validate `"read_databases"`, `"read_agents"`, and `"run_platform"` actions.
- `"DELETE"` was added to the `allow_methods` list in `CORSMiddleware`.

## Expected Effect
Unauthorized or cross-workspace access to database registries, agents, and platform sessions will return a `403 Forbidden` response. Frontend web applications will successfully complete `DELETE` requests.

## Validation
- Ran `pytest extraction/tests/test_api_endpoints.py`
- Ran `bash scripts/ci/run_basic_ci.sh`
All tests complete successfully, indicating no regressions.

## Risks
Endpoints now enforce workspace isolation and policy checks, which may surface permission errors if downstream clients rely on the previously unprotected behavior without valid credentials or authorization setup. However, the default workspace behavior mitigates backward compatibility risk for internal SDK flows.

---
*PR created automatically by Jules for task [18308365976890591370](https://jules.google.com/task/18308365976890591370) started by @tteon*

### DIFF
--- a/extraction/agent_server.py
+++ b/extraction/agent_server.py
@@ -792,28 +792,14 @@ async def platform_chat_send(request: PlatformChatRequest):
 
 @app.get(RuntimePath.PLATFORM_CHAT_SESSION, response_model=PlatformSessionResponse)
 @track("agent_server.platform_chat_session_get")
-async def platform_chat_session_get(
-    session_id: str,
-    workspace_id: str = Query(default="default", pattern=WORKSPACE_ID_PATTERN)
-):
-    try:
-        require_runtime_permission(role="user", action="run_platform", workspace_id=workspace_id)
-    except PermissionError as e:
-        raise HTTPException(status_code=403, detail=str(e))
+async def platform_chat_session_get(session_id: str):
     history = [PlatformTurn(**row) for row in platform_session_store.get(session_id)]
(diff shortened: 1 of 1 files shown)


tteon:
Review triage: not planning to merge this branch as-is. The only piece worth preserving is the `DELETE` CORS allowance; that should be ported into #21. The rest of this branch targets the legacy `extraction/agent_server` path and is stale relative to the current `runtime/` ownership.

PR #7, opened 2026-04-14
Fix Cypher injection vulnerability in schema manager

tteon:
**Severity:** High
**Vulnerability:** Unsafe dynamic Cypher/query construction leading to injection.
**Impact:** A malicious payload extracted from documentation (via LLM `records`) could insert arbitrary Cypher queries into `extraction/schema_manager.py` when it attempts to `session.run` queries for constraints and indexes.
**Fix:** Modified `apply_schema` to remove existing backticks from dynamically constructed schema identifiers (`label` and `prop_name`), and then explicitly wrap the values in backticks.
**Validation:** Verified the fix passes all unit tests and integration pipelines by executing `bash scripts/ci/run_basic_ci.sh` along with explicit linting of agent docs.
**Risks:** Low. Only impacts schema application which is now fully escaped.

---
*PR created automatically by Jules for task [6200813466811123728](https://jules.google.com/task/6200813466811123728) started by @tteon*

### DIFF
--- a/extraction/schema_manager.py
+++ b/extraction/schema_manager.py
@@ -45,31 +45,35 @@ class SchemaManager:
                     for prop_name, config in props.items():
                         # 1. Unique Constraints
                         if config.get('constraint') == 'UNIQUE':
-                            constraint_name = f"constraint_{label}_{prop_name}_unique"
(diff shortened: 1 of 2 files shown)


tteon:
Review triage: this is the preferred schema hardening PR in the injection-fix stream. The patch improves quoting/sanitization in `extraction/schema_manager.py`, but I would block merge on one follow-up: add a focused regression test covering labels/property names containing backticks or injection-shaped content. Once that test exists, this is the right PR to merge.

PR #11, opened 2026-04-17
perf: Migrate prompt history logging to append-only JSONL format

tteon:
### What
Migrates `prompt_history.json` to an append-only `prompt_history.jsonl` format. 
Updates `PromptManager.log_result` to open the file in append mode (`a`) and write a single JSON string per call, instead of loading the entire JSON file into memory, appending the list, and rewriting it.

### Why
Previously, `PromptManager.log_result` read and re-wrote the *entire* prompt history file every time a prompt was logged, which happens on every extraction and linking call. This created an obvious O(N²) / repeated file I/O bottleneck during dataset ingest workloads and was prone to race conditions. Switching to append-only JSONL eliminates this overhead and aligns with the repo's tracing standards.

### Expected Impact
Removes the O(N^2) I/O bottleneck in prompt history logging, leading to a bounded performance win during extraction/linking ingestion loops.

### Validation
- Added a new unit test `extraction/tests/test_prompt_manager.py` to verify `log_result` appends correct `.jsonl` lines.
- Ran `uv run pytest extraction/tests/test_prompt_manager.py` locally.
- Ran repo-local `bash scripts/ci/run_basic_ci.sh` to ensure no regressions in runtime shell contracts or other test suites.

### Risks
Tools or scripts relying exclusively on reading `prompt_history.json` as a single JSON array might break if they have not been migrated to handle the `.jsonl` format, although no such hard dependencies were found in the hot path.

---
*PR created automatically by Jules for task [4506299064274984786](https://jules.google.com/task/4506299064274984786) started by @tteon*


tteon:
Review triage: this is the preferred JSONL migration PR. It carries the same `PromptManager.log_result()` improvement as #12/#22 and, unlike them, includes a focused regression test in `extraction/tests/test_prompt_manager.py`. Before merge, please decide whether to also migrate the checked-in `extraction/prompt_history.json` payload or explicitly document the one-time format switch to `prompt_history.jsonl`. I would keep this one and close #12/#22 as superseded.

PR #12, opened 2026-04-17
perf: optimize PromptManager history logging with JSONL append

tteon:
## What
Changes `PromptManager.log_result()` to append log entries to an append-only `.jsonl` file (`prompt_history.jsonl`) instead of reading, decoding, appending, and rewriting the full `prompt_history.json` array on every call.

## Why
As the prompt history file grows, reading and parsing the entire JSON array on every logged prompt result creates an O(N^2) scaling bottleneck in terms of file I/O and memory overhead. Converting to an append-only `.jsonl` format changes this operation to an O(1) append.

## Expected Impact
A significant reduction in disk I/O and memory usage when running long extraction or reasoning workflows, eliminating the latency spikes caused by `PromptManager` serialization as history grows.

## Validation
- Converted `extraction/prompt_history.json` to `extraction/prompt_history.jsonl` maintaining data parity.
- Verified test suite passes using `bash scripts/ci/run_basic_ci.sh`.
- Local tests confirm successful appending of new `.jsonl` lines.

## Risks
- Tooling or external scripts that assumed `prompt_history.json` was a flat JSON array may fail to parse `prompt_history.jsonl`.
- If `PromptManager` writes are highly concurrent without a file lock, minor interleaving might theoretically happen, though this risk is unchanged from the original single-file write implementation.

---
*PR created automatically by Jules for task [637246340096042881](https://jules.google.com/task/637246340096042881) started by @tteon*


tteon:
Review triage: closing this in favor of #11. Same JSONL append idea, but #11 is the better merge candidate because it adds a focused `PromptManager` regression test. If anything from this branch is still wanted, it is only the checked-in history-file migration.

PR #105, opened 2026-05-23
fix: normalize multi-word FinDER categories to declared CATEGORIES form

### LINKED ISSUE #104: by_category and sample_per_category silently drop multi-word FinDER categories
## Required Information

**Describe the Bug:**

`seocho/eval/benchmarks/finder.py` declares the category taxonomy in the `CATEGORIES` constant using PascalCase (`CompanyOverview`, `ShareholderReturn`), but `_normalize_category` only strips whitespace and keeps the source casing. The raw HuggingFace labels are `Company overview` and `Shareholder return`, so normalization produces `Companyoverview` and `Shareholderreturn`, which never match the declared constants.

`by_category` checks the requested name against `CATEGORIES` and then filters on the normalized stored value, so both multi-word categories resolve to an empty slice. `sample_per_category` iterates over `CATEGORIES` and calls `by_category` for each, so it returns a "balanced" sample that is missing those two categories with no error. The six single-word categories are unaffected because there is no casing difference after the space strip.

**Steps to Reproduce:**

```python
from seocho.eval.benchmarks.finder import by_category, sample_per_category

len(by_category("CompanyOverview"))     # 0, expected 1081
len(by_category("ShareholderReturn"))   # 0, expected 490
len(by_category("Financials"))          # 990, fine (single word)

rows = sample_per_category(50)
sorted({r["category"] for r in rows})
# ['Accounting', 'Financials', 'Footnotes', 'Governance', 'Legal', 'Risk']
# CompanyOverview and ShareholderReturn are absent
```

**Expected Behavior:**

`by_category` resolves every name in `CATEGORIES` to a non-empty slice, and `sample_per_category` returns all eight categories. Category matching should be insensitive to spacing and casing, so `Company overview`, `COMPANY OVERVIEW`, and the already-normalized `CompanyOverview` all map to the single declared spelling.

**Observed Behavior:**

`by_category("CompanyOverview")` and `by_category("ShareholderReturn")` return 0 rows. `sample_per_category` drops both categories silently, so a benchmark run loses a quarter of the taxonomy without any signal.

**Environment Details:**

- seocho version: 0.4.1
- OS: macOS
- Python: 3.12

---

## Optional Information

The root cause is two sources of truth that disagree: the hardcoded `CATEGORIES` set and the algorithmic normalizer.

Patch ready:

- Derive a reverse lookup from `CATEGORIES` so the normalizer cannot drift from the declared set, with a title-cased fallback for unknown labels.
- Bump the cached parquet filename so a stale cache built under the old normalization is not reused, and force a fresh `.map` on a cache miss. HuggingFace fingerprints the mapped function by bytecode and does not see edits to its transitive `_normalize_category` dependency, so the arrow cache would otherwise re-emit the old schema.
- Add a load-time check that compares observed categories against `CATEGORIES` and warns on drift instead of failing silently.
- Regression tests covering all eight categories, the balanced sampler, and the data-versus-declaration check.

The duplicate loader at `examples/teaching/_shared/finder_loader.py` carries the same logic and likely has the same bug. It is out of scope for this fix.

1wos:
Fixes #104

## Feature

`by_category` and `sample_per_category` in `seocho/eval/benchmarks/finder.py` now resolve every declared category, including the two multi-word ones (`CompanyOverview`, `ShareholderReturn`).

## Why

`_normalize_category` stripped whitespace but kept the source casing, so the raw HuggingFace labels `Company overview` and `Shareholder return` became `Companyoverview` and `Shareholderreturn`. Those never matched the PascalCase entries in `CATEGORIES`, so:

- `by_category("CompanyOverview")` returned 0 rows (1081 expected)
- `by_category("ShareholderReturn")` returned 0 rows (490 expected)
- `sample_per_category` produced a "balanced" sample missing both categories, with no error

The six single-word categories were unaffected, which is why this stayed hidden. The root cause is two sources of truth that disagreed: the hardcoded `CATEGORIES` set and the algorithmic normalizer.

## Design

- Derive `_CANONICAL_BY_KEY` from `CATEGORIES` so the normalizer is built from the declared set and cannot drift from it. Lookup is casing- and spacing-insensitive; unknown labels fall back to a title-cased, space-stripped form so a future FinDER category does not hard-fail.
- Bump the parquet cache filename to `finder_corpus_v2.parquet` and pass `load_from_cache_file=False` on a cache miss. HuggingFace fingerprints `.map` by the mapped function's bytecode, which does not change when its transitive `_normalize_category` dependency is edited, so the stale arrow cache would otherwise re-emit the old schema.
- Add `_verify_categories` at load time to compare observed categories against `CATEGORIES` and warn on drift instead of silently producing empty slices. It warns rather than raising so a benchmark load degrades loudly without crashing.

## Expected Effect

All eight categories resolve to non-empty slices and the balanced sampler covers the full taxonomy.

## Validation

New `seocho/tests/test_finder_categories.py`:

- pure-function: normalization produces the declared form for multi-word labels, is idempotent, and every normalized label is in `CATEGORIES`
- dataset-backed: `by_category` is non-empty for all eight categories, `sample_per_category` covers all eight, and observed categories equal `CATEGORIES`

Local run: `52 passed` across `test_finder_categories.py`, `test_eval_benchmark.py`, `test_benchmarking.py`, `test_finder_benchmark_script.py` (clean cache rebuild). flake8 clean at the repo line length (88) for the changed lines.

## Risks

- Existing local caches named `finder_corpus.parquet` are left in place but no longer read; the new `v2` name triggers a one-time rebuild from HuggingFace.
- The duplicate loader at `examples/teaching/_shared/finder_loader.py` carries the same logic and likely the same bug. It is left out of scope to keep this change focused.


tteon:

Re-applied against current file locations in #555, with attribution to this PR. Closing in favour of that. Thank you for the fix — it was a real bug and the diagnosis held up on re-verification.

PR #111, opened 2026-05-24
perf: pre-compile explicit direction regex in cypher guard

tteon:
### What
Moved the dynamic `re.compile()` call for the explicit direction matcher in `seocho/query/guards.py` out of the `detect_wrong_direction` function and into a module-level constant (`_RE_EXPLICIT_DIRECTION`).

### Why
`detect_wrong_direction` is called iteratively over potentially many queries or nodes in a hot path. Recompiling the regex on every call is unnecessary overhead. Pre-compiling it as a module-level constant prevents redundant loop work.

### Expected Impact
Qualitative only. Slight reduction in CPU overhead during Cypher validation guards due to cached regex matching.

### Validation
Ran `uv run pytest seocho/query/` and `uv run pytest seocho/tests/test_cypher_builder.py` locally and they passed. Basic CI passes. 

### Risks
Minimal. Python's `re.Pattern` objects are thread-safe and reusable. No functional behavior was changed.

---
*PR created automatically by Jules for task [6027062815467602704](https://jules.google.com/task/6027062815467602704) started by @tteon*

### DIFF
--- a/docs/QUICKSTART.md
+++ b/docs/QUICKSTART.md
@@ -1,0 +1,67 @@
+# SEOCHO Quickstart
+
+Fastest GitHub-first path to one successful SEOCHO run.
+
+For the full runtime deployment guide (Docker stack, services, env), see
+[docs/RUNTIME_DEPLOYMENT.md](docs/RUNTIME_DEPLOYMENT.md).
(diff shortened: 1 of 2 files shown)


tteon:

PR #113, opened 2026-05-25
fix: carry prompt body and semantic tags on extraction and query trace spans

### LINKED ISSUE #106: Opik traces split semantic context and prompt body across separate spans
## Required Information

**Describe the Request:**

When `SEOCHO_TRACE_BACKEND=opik`, no single Opik span carries both a meaningful name/tags and the actual LLM prompt and completion. The two span sources each cover only half of what is needed to audit extraction and query quality:

- `log_extraction` (`seocho/index/pipeline.py:945`) emits a `sdk.extraction` span with `text_preview` (200 chars of the source text), ontology name, model, node/relationship counts, score, and elapsed time. It does not include the rendered system/user prompt or the model's JSON completion.
- `log_query` (`seocho/local_engine.py:788`) emits a `sdk.query` span with the question, a `cypher_preview`, and counts. It does not include the full prompt or the synthesized answer.
- `track_openai` (`seocho/store/llm.py:251`) wraps the OpenAI-compatible client, so every LLM call also produces a `chat_completion_create` span that does contain the full prompt and completion. But that span has a generic name, no tags, and no link to the stage (extraction vs linking vs query), the ontology, or `workspace_id`.

The result: in the Opik UI you can see counts with a stage label but no prompt (`sdk.extraction`), or the full prompt with no stage label (`chat_completion_create`). There is no correlated, filterable view of "what prompt produced this extraction for this ontology in this workspace."

**Steps to Reproduce:**

1. Set `SEOCHO_TRACE_BACKEND=opik` and provider keys, then index a document and run a query.
2. Open the Opik project.
3. Observe `sdk.extraction` / `sdk.query` spans (counts, no prompt body) alongside `chat_completion_create` spans (prompt body, generic name, no tags).
4. Try to filter the trace list by stage (extraction vs linking vs query), by ontology, or by `workspace_id`. None of these are available as tags or span names on the spans that carry the prompt.

**Expected Behavior:**

A single span per LLM call that carries both the semantic context and the payload, so traces are auditable and filterable. Concretely, one of:

- enrich `sdk.extraction` / `sdk.query` with the rendered system and user prompt and the model completion, and tag them with stage, ontology name, and `workspace_id`; or
- name and tag the auto-wrapped client spans per stage (`seocho.index.extract`, `seocho.index.link`, `seocho.query.text2cypher`) and stamp `workspace_id` and ontology metadata on them.

This lines up with the observability intent in `CLAUDE.md` section 9 (include `workspace_id` and user context in trace metadata) and section 14 (traces should preserve enough metadata to audit routing, disambiguation, and synthesis).

**Observed Behavior:**

Prompt and completion live only on anonymous `chat_completion_create` spans. The semantically named `sdk.extraction` / `sdk.query` spans omit them. Neither set carries `workspace_id`, ontology, or stage tags, so the Opik traces cannot be filtered by tenant, schema, or pipeline stage.

**Environment Details:**

- seocho version: 0.4.1
- OS: macOS
- Python: 3.12
- Opik: cloud (`SEOCHO_TRACE_BACKEND=opik`, `SEOCHO_TRACE_OPIK_MODE=hosted`)

---

## Optional Information

This is a request only. Filing now to track the gap; the implementation can come later. No patch is attached.

A workaround today is to call `seocho.tracing.log_span(name=..., input_data={system, user}, output_data=..., tags=[...], metadata={workspace_id, ontology, stage})` directly around each LLM call, which produces named, tagged spans that include the prompt. That works at the caller level but is not part of the SDK's default tracing path.

1wos:
Closes #106

## Feature

`log_extraction` and `log_query` now emit one `sdk.extraction` / `sdk.query` span that carries the rendered prompt and the model completion (or synthesized answer) next to the semantic context, tagged by stage, ontology, workspace, and a provider-qualified model.

## Why

With `SEOCHO_TRACE_BACKEND=opik` the prompt body and the semantic context lived on different spans:

- `sdk.extraction` / `sdk.query` carried counts and a fixed `["extraction"|"query", "model:..."]` tag set, but not the prompt or completion.
- the prompt and completion appeared only on the auto-wrapped `chat_completion_create` span, which has a generic name and no stage, ontology, or workspace tags.
- the `model:` tag used the bare model id (`kimi-k2.5`), dropping the provider.

So no single span could be filtered by tenant (`workspace_id`), schema (`ontology`), or pipeline stage, and the counts and the prompt could not be correlated.

## Design

- `log_extraction` / `log_query` take optional `system_prompt`, `user_prompt`, `completion` (or `answer`), `workspace_id`, and `provider`. The prompt goes in `input_data`, the completion or answer in `output_data`, and the span gains `stage:`, `ontology:`, `workspace:` and `model:{provider}/{model}` tags.
- `CanonicalExtractionEngine` retains the last rendered prompt and completion (`last_system` / `last_user` / `last_completion`) so the indexing pipeline can attach them without changing the extractor's return contract.
- All new parameters are optional. Callers that omit them keep the previous output (bare `model:` tag, no prompt or completion), so the change is backward compatible.

## Expected Effect

A single span is enough to audit an extraction or query: the prompt, the completion, and the stage / ontology / workspace / model that produced it are all filterable in the Opik UI.

## Validation

New tests in `seocho/tests/test_tracing.py`:

- the extraction span carries `system_prompt`, `user_prompt`, `completion`, and the `stage` / `ontology` / `workspace` / `model:{provider}/{model}` tags
- the query span carries the `answer` and the same semantic tags
- backward compatible: with no `provider` the model tag stays unqualified and the prompt and completion are omitted

Local run: `56 passed` across `test_tracing.py`, `test_observability_degradation.py`, `test_llm_backends.py`, `test_extraction_engine.py`, `test_indexing.py`, `test_observability.py`. End to end through the real indexing path with four providers (kimi, openai, grok, deepseek): each `sdk.extraction` span carried the prompt and a correct `model:{provider}/{model}` tag. flake8 clean at the repo line length (88) for the changed lines.

## Risks

- `log_extraction` reports the most recent prompt and completion only, so on a multi-chunk document the span shows the last chunk's prompt; the counts stay aggregate.
- The query path attaches the synthesized answer but not the text2cypher prompt, which can be a follow-up.
- No change to the auto-wrapped `chat_completion_create` span; this resolves the gap on the `sdk.*` spans (Option 1 in the issue) rather than renaming and tagging the client spans.


tteon (approved):

tteon:

Re-applied against current file locations in #555, with attribution to this PR. Closing in favour of that. Thank you for the fix — it was a real bug and the diagnosis held up on re-verification.

PR #151, opened 2026-05-28
[DRAFT] perf: fix N+1 query overhead in ontology coverage stats

tteon:
**What:** 
Updated `Ontology.coverage_stats()` in `seocho/ontology.py` to use `UNION ALL` to batch Cypher count queries into chunks of 50, rather than issuing separate N+1 queries for every single node label and relationship type defined in the ontology. Updated `FakeGraphStore` in tests to support this query pattern.

**Why:** 
For very large ontologies, loading schema counts creates an N+1 query problem that repeatedly hits the graph database for individual `MATCH (n:\`Label\`) RETURN count(n)` calls, causing severe and unnecessary network round-trip overhead.

**Expected Impact:** 
Significant reduction in wait times (latency) when executing schema/coverage queries against live graph instances with large ontology definitions.

**Validation:** 
- Updated `FakeGraphStore` in test suite to correctly parse `UNION ALL` chunks.
- Verified test suite passes via `uv run pytest seocho/tests/test_ontology.py` and `seocho/tests/test_ontology_artifact_promotion.py`.
- Ran full repository CI (`bash scripts/ci/run_basic_ci.sh`) and confirmed everything passed without issue.

**Risks:** 
- If an individual query within a batched `UNION ALL` group were to fail due to a syntax issue (e.g., malformed label escaping), the entire batch (up to 50 types) would fail and report 0 counts. To mitigate this, label sanitization (stripping newlines and backticks) is applied prior to construction.

*(Draft PR as requested by constraints)*

---
*PR created automatically by Jules for task [2117800069781823317](https://jules.google.com/task/2117800069781823317) started by @tteon*

### DIFF
--- a/seocho/ontology.py
+++ b/seocho/ontology.py
@@ -2356,15 +2356,27 @@ class Ontology:
         total_defined_nodes = len(self.nodes)
         populated_nodes = 0
 
-        for label in self.nodes:
+        # Batch nodes counting
+        nodes_list = list(self.nodes)
+        node_counts = {lbl: 0 for lbl in nodes_list}
+        chunk_size = 50
+        for i in range(0, len(nodes_list), chunk_size):
+            chunk = nodes_list[i : i + chunk_size]
+            queries = []
+            for lbl in chunk:
+                safe_lbl = lbl.replace('\n', '').replace('\r', '').replace('`', '')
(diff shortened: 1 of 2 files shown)


tteon:

PR #154, opened 2026-05-29
 fix: migrate deprecated id() to elementId() in GDS analytics projection

### LINKED ISSUE #153: bug: deprecated id() still used in GDS   analytics projection queries
## Description

  Two locations in the codebase still use the
  deprecated `id()` Cypher function
  instead of `elementId()`, violating the
  project's own policy documented in
  **ADR-0020** and **CLAUDE.md §8**.

  ## Affected Files

  | File | Lines | Code |
  |------|-------|------|
  | `seocho/agent/graph_loop.py` | 211, 215–216
  | Default `analytics_projection` node/rel
  queries |
  | `seocho/gds/__init__.py` | 27–29 | Module
  docstring example |

  ## Problem

  `id(n)` returns an internal integer that
  DozerDB/Neo4j is deprecating.
  It can change across restarts or data
  migrations, making it an unstable
  identifier. `elementId(n)` is the stable,
  durable successor.

  The rest of the codebase already uses
  `elementId()` consistently
  (e.g. `semantic_agents.py`, `http_runtime.py`,
   `cypher_builder.py`),
  making these two locations the remaining gaps.

  ## Expected Behavior

  All runtime and query-time Cypher paths should
   use `elementId()` per
  ADR-0020 and CLAUDE.md §8.

  ## References

  - ADR-0020: P1 ElementId Migration, Health
  Split, and Readiness Fallback
  - CLAUDE.md §8: *"prefer `elementId(...)` over
   deprecated `id(...)` in
    query-time/runtime Cypher paths"*

ajm6238:
Closes #153

## Summary

  - Replace deprecated `id(e)` with
  `elementId(e)` in the default GDS Cypher
  projection queries in
  `seocho/agent/graph_loop.py`
  - Update the matching docstring example in
  `seocho/gds/__init__.py` to reflect the same
  fix
  - `WHERE elementId(e1) < elementId(e2)`
  preserves the original deduplication semantics
   (preventing double-counting of undirected
  pairs) using lexicographic ordering instead of
   numeric

  ## Motivation

  Per **ADR-0020** and **CLAUDE.md §8**, `id()`
  is deprecated in DozerDB/Neo4j and should be
  replaced with `elementId()` across all
  query-time/runtime Cypher paths. The rest of
  the codebase (`semantic_agents.py`,
  `http_runtime.py`, `cypher_builder.py`, etc.)
  already uses `elementId()` — these two
  locations were the remaining gaps.

  ## Files Changed

  | File | Lines | Change |
  |------|-------|--------|
  | `seocho/agent/graph_loop.py` | 211, 215–216
  | `id(e)` → `elementId(e)` in default
  analytics projection |
  | `seocho/gds/__init__.py` | 27–29 | Same fix
  in module docstring example |

  ## Test plan

  - [ ] Verify `grep -r "RETURN id(" seocho/`
  returns no matches after this change
  - [ ] Run existing agent engine tests: `pytest
   seocho/tests/test_agent_engine.py`
  - [ ] If a DozerDB instance is available,
  confirm GDS projection succeeds with
  `elementId()`-based queries


tteon:
Holding this one rather than porting it with the other 20 contributor fixes (see #548, #549, #551, #552, #553, #555).

The deprecated `id()` usage you identified is real and still present at `src/seocho/agent/graph_loop.py:230-240`. The concern is where the replacement lands: both edited queries are the `node_query`/`rel_query` fed to `gds.graph.project.cypher`, the legacy Cypher-projection procedure, which historically requires the projection query's `id` column to be Neo4j's **native integer** id rather than a string.

This codebase's own convention points the same way — `src/seocho/gds/__init__.py:282` converts back with `elementId(gds.util.asNode(nodeId))` only *after* an algorithm runs, i.e. `elementId()` is used downstream of the projection and never inside it.

So this needs verification against a live DozerDB/GDS instance before landing, or scoping down to exclude the two projection queries specifically. Not rejected — just the one PR in the batch whose fix could itself be a regression, and it deserves a real test rather than a guess.

Tracking it so it doesn't get lost.

tteon:
Backlog review: keeping this draft. The branch predates the src/ layout, and changing id() to elementId() changes numeric projection identities into strings while retaining the legacy GDS Cypher projection API. This needs a coherent projection migration, not a general deprecation substitution. Extract only the intended change onto current main and verify projection creation, algorithms, and result-to-node lookup against the supported live GDS version. The legacy API and migration are documented at https://neo4j.com/docs/graph-data-science/current/management-ops/graph-creation/graph-project-cypher-legacy/ . No live GDS validation was run in this review.

PR #167, opened 2026-05-31
perf: batch graph count queries via UNION ALL

tteon:
What:
Batched individual node and relationship counting queries during schema validation and metadata population using Cypher's `UNION ALL` feature, chunking them in groups of 50.

Why:
Previously, the system would issue an independent database query (e.g. `MATCH (n:\`Label\`) RETURN count(n) AS cnt`) for every single node label and relationship type defined in an ontology or graph schema. For complex ontologies, this N+1 querying approach introduced massive linear overhead due to database round-trips.

Expected Impact:
Significantly faster query latency when analyzing ontology coverage and graph size against Neo4j and DozerDB databases, reducing O(N) round-trips to O(N/50).

Validation:
- Tested with dummy batch implementations
- Ran `bash scripts/ci/run_basic_ci.sh` with 100% pass rate

Risks:
- Some graph databases strictly fail queries referencing undefined node or edge tables. We included a fallback to individual queries if the batched chunk throws an exception, mitigating this risk.

---
*PR created automatically by Jules for task [18164767895162284909](https://jules.google.com/task/18164767895162284909) started by @tteon*

### DIFF
--- a/seocho/ontology.py
+++ b/seocho/ontology.py
@@ -2403,35 +2403,85 @@ class Ontology:
         total_defined_nodes = len(self.nodes)
         populated_nodes = 0
 
-        for label in self.nodes:
+        nodes_list = list(self.nodes)
+        for i in range(0, len(nodes_list), 50):
+            chunk = nodes_list[i:i+50]
+            if not chunk: continue
+
+            queries = []
+            for label in chunk:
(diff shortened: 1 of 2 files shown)


tteon:

PR #170, opened 2026-06-01
fix: read_json/jsonl warn instead of silently dropping non-object items

### LINKED ISSUE #169: bug: read_json_file/read_jsonl_file drop non-object array items silently
## Summary

`seocho.index.file_reader.read_json_file` and `read_jsonl_file` expect an array
of objects (resp. one JSON object per line) with a `content` field. Any array
item / line that is **not** an object is dropped **silently** — no exception, no
log line.

This is inconsistent with the functions' own behavior: `read_jsonl_file` already
emits a `logger.warning` for malformed (unparseable) lines, but a line that
parses to a valid-but-non-object value (a string, number, or `null`) vanishes
with no signal. It also conflicts with the project's robustness principle in
`CLAUDE.md` §18 ("무성(silent) fallback 금지 — 실패는 실패로 기록").

The practical symptom: a JSON file that is an array of plain strings, e.g.

```json
["first document text", "second document text"]
```

returns **zero records**, and the user has no way to tell whether the file was
empty, malformed, or simply the wrong shape.

## Reproduction

```python
import json, tempfile
from pathlib import Path
import seocho.index.file_reader as fr

with tempfile.TemporaryDirectory() as d:
    p = Path(d) / "docs.json"
    p.write_text(json.dumps(["first document text", "second document text"]))
    print("json  ->", len(fr.read_json_file(p)))   # 0, no warning

    p2 = Path(d) / "docs.jsonl"
    p2.write_text(json.dumps("line one") + "\n" + json.dumps("line two") + "\n")
    print("jsonl ->", len(fr.read_jsonl_file(p2)))  # 0, no warning
```

Output (note: completely silent):
```
json  -> 0
jsonl -> 0
```

## Expected

The object-only contract is fine to keep — these functions are documented to
read objects with a `content` field. But a skipped item should be **visible**,
not silent: emit a `logger.warning` (naming the file, index/line, and item type)
when a non-object item is skipped, mirroring the existing malformed-line warning.

That way the documented shape stays authoritative, while a mismatched input
(string array, stray scalar) surfaces a diagnosable warning instead of an
unexplained empty result.

## Affected code

- `seocho/index/file_reader.py` — `read_json_file` (`if isinstance(item, dict)` branch)
- `seocho/index/file_reader.py` — `read_jsonl_file` (same)

## Proposed fix

Skip non-object items **with a `logger.warning`** instead of dropping them
silently. PR with regression tests incoming.

JunHyeok-Cha:
Closes #169

## Problem

`read_json_file` and `read_jsonl_file` in `seocho/index/file_reader.py` expect an
array of objects (resp. one JSON object per line) with a `content` field, but any
array item / line that is **not** an object is dropped **silently** — no
exception, no log line.

This is inconsistent with the functions' own behavior (`read_jsonl_file` already
warns on malformed, unparseable lines) and with the robustness principle in
`CLAUDE.md` §18 ("무성(silent) fallback 금지"). Symptom: a JSON file that is an
array of plain strings returns zero records, and the user can't tell whether the
file was empty, malformed, or simply the wrong shape.

## Reproduction (before this PR)

```python
import json, tempfile
from pathlib import Path
import seocho.index.file_reader as fr

with tempfile.TemporaryDirectory() as d:
    p = Path(d) / "docs.json"
    p.write_text(json.dumps(["first document text", "second document text"]))
    print(len(fr.read_json_file(p)))   # 0, no warning
```

## Fix

Keep the documented object-only contract, but emit a `logger.warning` (naming the
file, index/line, and item type) when a non-object item is skipped — mirroring the
existing malformed-line warning. No behavior change for valid object inputs; a
mismatched input shape now surfaces a diagnosable warning instead of an
unexplained empty result.

## Tests

Added regression tests in `seocho/tests/test_file_indexer.py` asserting both the
skip and the warning:
- `test_json_array_of_non_objects_skipped_with_warning`
- `test_json_array_mixed_keeps_objects`
- `test_jsonl_non_object_lines_skipped_with_warning`

```
$ python -m pytest seocho/tests/test_file_indexer.py -q
19 passed
```


tteon:
Triage vs current main: the asymmetry this fixes is still live (read_json_file/read_jsonl_file silently skip non-dict items while malformed lines warn), but the file moved to src/seocho/index/file_reader.py and both readers were rewritten (streaming open(), _record_metadata helper), so the hunks won't apply verbatim. The fix re-applies as two ~3-line isinstance guards mirroring the existing malformed-line warning; the tests port directly to tests/seocho/test_file_indexer.py. Easiest of the stale externals to revive — a rebase (or permission to push a rebase onto your branch) gets this landed.

tteon:
The reviewed change is now incorporated in merged #673. Older-path/history proposals were ported onto current main, with current metadata/structured-output behavior preserved and missing CLI/environment contracts covered. Original contribution retained through this PR reference; closing as incorporated.

PR #192, opened 2026-06-06
feat(bench/finder_judge): skip LLM judge on empty candidates

Min-hye-Kim:
## Problem
`judge_one()` sent every candidate to the judge, including empty ones. An empty answer is `incorrect` by definition, but the judge was still asked to rule on it, nd when the generator and the judge are the same vendor it **agrees with itself**:
38 empty-candidate cells came back `correct` or `partial`.          
A cross-vendor recheck of those same 38 cells (DeepSeek → `gpt-4o-mini`) returned `incorrect` on all 38, so the verdicts were judge artefacts rather than scores. Each one also cost a model call.    
<br/>

## Fix

- Return `incorrect` / `score=0.0` with `sanity_skipped=True` at the top of   `judge_one()` when the candidate is empty or whitespace-only, before any model call.
- `_safe_str` already normalises `None` and `NaN` to `""`, so null answers in the   saved partial JSONs take the same branch - no separate null check.

The guard is deliberately empty-only. An earlier revision of this PR auto-failed candidates under 3 meaningful tokens ; per review that was too aggressive, and the `<1 meaningful token` alternative does not fix it either - the tokenizer splits
`"8.4%"` on the `.`, leaving `8` and `4`, both dropped as single characters, so the count is 0 under either rule.         
Terse-but-correct FinDER numerics reach the judge unchanged.    
<br/>    

## Tests (`test_finder_judge.py`)

- `""`, `"   "`, `"\n\t "`, `None` and `NaN` return `incorrect` with `sanity_skipped=True`; the stub judge raises `AssertionError` if it is reached at all.
- `"8.4%"`, `"$5B"`, `"12.7"` and `"Apple Inc."` still reach the judge rather than   being auto-failed.

`pytest tests/seocho/test_finder_judge.py tests/seocho/test_finder_eval_helpers.py` → 28 passed (both modules load `finder_judge.py`). The two pre-existing `judge_one` wiring tests are byte-identical to `main`.     
<br/>

### DIFF
--- a/scripts/benchmarks/finder_judge.py
+++ b/scripts/benchmarks/finder_judge.py
@@ -115,6 +115,22 @@ def _parse_judge(text: str) -> dict:
 
 
 def judge_one(llm, query: str, gold: str, candidate: str) -> dict:
+    # Sanity guard - an empty candidate is incorrect by definition, so skip
+    # the judge call. LLM judges (especially same-vendor as the generator)
+    # tend to return a plausible-sounding correct/partial on empty answers; a
+    # cross-vendor recheck of 38 such cells flipped all 38 to incorrect.
+    # Deliberately empty-only: terse-but-correct answers ("8.4%", "$5B") must
+    # still be judged, never auto-failed.
+    if not _safe_str(candidate).strip():
+        return {
+            "verdict": "incorrect",
+            "score": 0.0,
+            "rationale": "empty candidate answer (sanity guard, no judge call)",
+            "matched": [],
+            "missing_or_wrong": [],
+            "parse_error": False,
+            "sanity_skipped": True,
+        }
     user = (f"QUESTION:\n{_safe_str(query)}\n\n"
             f"GOLD ANSWER (ground truth):\n{_safe_str(gold)}\n\n"
             f"CANDIDATE ANSWER:\n{_safe_str(candidate)}")
--- a/tests/seocho/test_finder_judge.py
+++ b/tests/seocho/test_finder_judge.py
@@ -9,6 +9,8 @@ from __future__ import annotations
 import importlib.util
 from pathlib import Path
 
+import pytest
+
 _BENCH = Path(__file__).resolve().parents[2] / "scripts" / "benchmarks"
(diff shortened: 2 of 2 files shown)


tteon:
Triage vs current main (the repo moved ~600 commits since this was opened): still valuable and near-clean to rebase — judge_one() and both tests are byte-compatible, and no empty-candidate guard exists on main yet. One ask before merge: the <3-meaningful-tokens auto-fail would score terse-but-correct FinDER numerics like "8.4%" or "$5B" as incorrect without a judge call — please tighten the guard to truly empty/whitespace-only candidates (or <1 meaningful token). Happy to land it right after that (a maintainer also needs to approve CI for fork PRs).

Min-hye-Kim:
@tteon     
Rebased onto the latest upstream/main and tightened the guard to truly empty/whitespace-only candidates, and below are some additional explanations regarding these changes.   
<br/>   

**Reasons for not selecting the `<1 meaningful token` option** :<br/>
- I didn't take `<1 meaningful token` option, because it still auto fails `"8.4%"`: the tokenizer splits on the `.`, leaving `8` and `4`, and both are dropped as single characters - so, the count is 0 there too.        
- `"$5B"` does survive at `<1` but `"8.4%"` does not, so only the empty / whitespace check clears both of the examples you mentioned..      
- That let me drop the token-count helper entirely, which in turn means the two pre-existing `judge_one` wiring tests no longer need editing. (they are byte-identical to main)    
<br/>   
 
**Validation** :
- focused tests: 28 passed (`test_finder_judge.py`, `test_finder_eval_helpers.py`)
- new cases pin both directions: empty / whitespace / `None` / `NaN` must not  reach the judge, and `"8.4%"`, `"$5B"`, `"12.7"`, `"Apple Inc."` must still be judged
<br/>

tteon:
The reviewed change is now incorporated in merged #673. Older-path/history proposals were ported onto current main, with current metadata/structured-output behavior preserved and missing CLI/environment contracts covered. Original contribution retained through this PR reference; closing as incorporated.

PR #211, opened 2026-06-07
fix: declare PyYAML for ontology imports

### LINKED ISSUE #137: pip install seocho then from seocho import Ontology fails with ModuleNotFoundError: yaml (PyYAML is not a core dependency)
## Required Information

**Describe the Bug:**

`seocho/ontology.py:50` imports `yaml` at module top level, but PyYAML is not in the core `dependencies` of `pyproject.toml` (only `requests` and `python-dotenv` are; PyYAML lives in the `local`/`ci`/`dev` extras). Because `seocho.store.__init__` -> `seocho/store/graph.py:30` runs `from seocho.ontology import Ontology`, importing the primary public API after a plain `pip install seocho` raises `ModuleNotFoundError: No module named 'yaml'`. This breaks `seocho.Ontology`, `OpenAIBackend`, `create_llm_backend`, `FAISSVectorStore`, and the Neo4j/Ladybug graph stores in a core install. `import seocho` itself succeeds (lazy `__getattr__`), so the failure only surfaces when a user touches the documented API. The `[ontology]` and `[embedded]` extras also omit pyyaml, so even `pip install seocho[ontology]` cannot load `Ontology`.

**Steps to Reproduce:**

1. `pip install seocho` into a clean environment.
2. `python -c "from seocho import Ontology"`.
3. `ModuleNotFoundError: No module named 'yaml'`.

**Expected Behavior:**

The core public API imports successfully after a plain `pip install seocho`.

**Observed Behavior:**

`Ontology` and the LLM / vector / graph stores fail to import unless an extra that happens to include PyYAML is installed.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

Fix either by adding `pyyaml` to core `dependencies`, or by moving `import yaml` inside the loader functions in `ontology.py` (and likewise in `agent_design.py:7` and `indexing_design.py:7`). The `[ontology]`/`[embedded]` extras should also include pyyaml.

Suggested labels: bug

szx19970521:
## Summary
- add `pyyaml` to core dependencies because `seocho.ontology` imports `yaml` at module import time
- add `pyyaml` to the `ontology` and `embedded` extras so those optional install paths can also import the documented API

Fixes #137.

## Verification
Before the change, a clean install of `seocho==0.4.1` fails:

```text
python -c "from seocho import Ontology; print('IMPORT_OK')"
ModuleNotFoundError: No module named 'yaml'
```

After installing this branch in a clean virtualenv:

```text
python -c "from seocho import Ontology; print('IMPORT_OK', Ontology.__name__)"
IMPORT_OK Ontology
```


tteon:
Closing: already fixed on main — `"pyyaml"` is in the core `[project] dependencies` block of `pyproject.toml`. Thanks for catching it; it landed via another route.

PR #223, opened 2026-06-08
feat(runtime): append-only audit log of access decisions (seocho-bk0)

tteon:
## What
Adds `runtime/audit.py` — an opt-in, append-only JSONL audit log of every access decision, wired into the policy enforcement point.

> Stacked on #222 → #221. Base is `feat/rbac-workspace-authz`; retarget to main after the parents merge.

## Why
The runtime had request logs (method/path/elapsed) but **no who-did-what audit trail**, and didn't capture the principal at all. Regulated-data enterprises require an immutable record of who did what, where, and whether it was allowed.

## How
- Vendor-neutral, **opt-in** sink mirroring the tracing contract: writes JSON Lines to `SEOCHO_AUDIT_LOG_PATH`. **Unset (default) → no-op**, so existing deployments are unaffected.
- One event per decision: `ts / request_id / subject / role / authenticated / workspace_id / action / decision(allow|deny) / reason`.
- Wired into the **single enforcement point** in `require_runtime_permission`: both the auth-off (workspace-format) path and the authenticated (role + ownership) path funnel through one decision that is recorded before any `PermissionError` is raised — so **allow and deny are both captured**.
- Write failures are logged, **never raised** — auditing must not take down the request path. `build_event` is pure for testability.

## Tests
`test_audit.py`: allow/deny event schema, default no-op when unconfigured, write+append to JSONL, write-failure swallowed, and an end-to-end integration asserting allow/deny rows land for viewer/user principals through `require_runtime_permission`. `run_basic_ci.sh` green (425).

### DIFF
--- a/extraction/tests/test_audit.py
+++ b/extraction/tests/test_audit.py
@@ -1,0 +1,80 @@
+"""Tests for the access-decision audit log (seocho-bk0)."""
+
+import json
+
+import pytest
+
+from runtime.audit import build_event, record_access
+from runtime.identity import ANONYMOUS, Principal
+from runtime.policy import require_runtime_permission
+
+USER = Principal(subject="alice", role="user", workspace_id="acme", authenticated=True)
+
+
+def _read(path):
+    with open(path, encoding="utf-8") as fh:
+        return [json.loads(line) for line in fh if line.strip()]
+
+
+def test_build_event_allow_and_deny():
+    allow = build_event(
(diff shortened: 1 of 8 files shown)


tteon:
Superseded by #228 (consolidated auth stack — rbac+audit+isolation cherry-picked cleanly onto main after the stacked-PR base deletion broke this one).

PR #224, opened 2026-06-08
feat(query): enforce tenant isolation at the query boundary via flag (seocho-a6d)

tteon:
## What
Wires the store's dormant `enforce_workspace_filter` safety net to the runtime query boundary (`QueryProxy`) behind `SEOCHO_ENFORCE_WORKSPACE_FILTER`, plus a fitness-function test.

> Stacked on #223 → #222 → #221. Base is `feat/access-audit-log`; retarget to main after parents merge.

## Why
`enforce_workspace_filter` existed on the store but **no caller ever passed it** — tenant isolation was opt-in and nobody opted in. This makes it a single deployment flag instead of a dead per-call argument.

## How (strangler-safe)
- `QueryProxy(enforce_workspace_filter=None)` → env default (`SEOCHO_ENFORCE_WORKSPACE_FILTER`, **off** by default). When on, it threads `workspace_id` + `enforce_workspace_filter=True` so any query whose Cypher doesn't scope to `$workspace_id` is refused.
- The new kwargs are passed **only when enforcement is on**, so the default path keeps the exact narrow call shape backends/test-doubles already accept → **zero behaviour change by default, no test breakage**.

## Fitness function
`test_query_proxy_workspace_enforcement.py`: default-off uses the narrow signature (guards against kwarg leakage), enforcement threads workspace_id + flag, refuses unscoped Cypher, env default honored.

## Scope note (honest)
**Flipping the default ON repo-wide is deliberately NOT done here** — it needs a live-DB pass auditing every runtime query template to reference `$workspace_id`, which can't be validated offline. The IDOR is already closed at the authorization layer by the workspace-ownership check in #222; this is **defense-in-depth** at the data plane, available via one flag. `run_basic_ci.sh` green (429).

### DIFF
--- a/extraction/tests/test_audit.py
+++ b/extraction/tests/test_audit.py
@@ -1,0 +1,80 @@
+"""Tests for the access-decision audit log (seocho-bk0)."""
+
+import json
+
+import pytest
+
+from runtime.audit import build_event, record_access
+from runtime.identity import ANONYMOUS, Principal
+from runtime.policy import require_runtime_permission
+
+USER = Principal(subject="alice", role="user", workspace_id="acme", authenticated=True)
+
+
+def _read(path):
+    with open(path, encoding="utf-8") as fh:
+        return [json.loads(line) for line in fh if line.strip()]
+
+
+def test_build_event_allow_and_deny():
+    allow = build_event(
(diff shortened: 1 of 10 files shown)


tteon:
Superseded by #228 (consolidated auth stack).

PR #234, opened 2026-06-09
fix(embed): batch embedding requests under the provider input cap

### LINKED ISSUE #127: EmbeddingBackend.embed sends all texts in one request with no batching, so a large add_batch fails entirely
## Required Information

**Describe the Bug:**

`embed()` issues a single request with `input=list(texts)` (`seocho/store/llm.py:634`) and no chunking. OpenAI-compatible embedding endpoints cap inputs per request (around 2048 items and a token ceiling), so a large `add_batch` builds one oversized request that returns a 400 and loses the whole batch instead of degrading.

**Steps to Reproduce:**

1. Index a batch that produces several thousand chunks with an OpenAI-compatible embedding model.
2. `embed()` sends them all in one request.
3. The provider returns 400 and the entire batch is lost.

**Expected Behavior:**

`embed()` splits inputs into provider-safe sub-batches and concatenates results.

**Observed Behavior:**

One request for all inputs; a 400 on large inputs loses the whole batch.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## What

`embed()` issued a single `/embeddings` request with all texts, so a large `add_batch` built one oversized request that 400s and lost the whole batch.

This chunks inputs into sub-batches of 2048 (the OpenAI-compatible per-request cap) and concatenates results, reordering each batch by the response `index` so output stays aligned with input. Applied to both `OpenAICompatibleBackend.embed` and `OpenAICompatibleEmbeddingBackend.embed` via a shared `_embed_in_batches` helper.

## Test

`test_openai_embedding_backend_batches_large_inputs`: 4101 inputs now produce three calls (2048 / 2048 / 5), nothing dropped, no call exceeds the cap. Embedding/vector suites pass (42).

Closes #127


tteon:

PR #235, opened 2026-06-09
fix(vector): normalize LanceDB scores to similarity so backends are comparable

### LINKED ISSUE #126: VectorStore score is FAISS inner-product (higher better) but LanceDB L2 distance (lower better), so scores are not comparable across backends
## Required Information

**Describe the Bug:**

`VectorStore` exposes one `search()` returning a `score`, but the two backends define it oppositely. `FAISSVectorStore` uses `IndexFlatIP` and returns inner-product similarity (higher is better). `LanceDBVectorStore.search` surfaces `row.get("_distance", ...)` as `score` (`seocho/store/vector.py:310`+), an L2 distance (lower is better). Code that ranks or thresholds on `score` through the shared interface inverts results when the backend changes.

**Steps to Reproduce:**

1. Run the same query through the FAISS and the LanceDB backend.
2. Sort results by `score`.
3. The orderings are reversed relative to true relevance.

**Expected Behavior:**

A single documented score convention across backends (e.g. both higher-is-better similarity), or an explicit `metric` field on `VectorSearchResult`.

**Observed Behavior:**

FAISS `score` is a similarity, LanceDB `score` is a distance; same API, opposite meaning.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## What

`VectorSearchResult.score` was FAISS inner-product (higher=better) for one backend but LanceDB **L2 distance** (lower=better) for the other, so the same field meant opposite things. Any `score.sort(reverse=True)` consumer (e.g. `ontology_control_plane.py:313`, `agent/matchmaker.py:125`) silently ranked LanceDB results **worst-first**.

This converts LanceDB distance to a similarity via `1/(1+distance)` (higher=better, bounded `(0, 1]`), leaving FAISS unchanged, and adds a `metric` field documenting the convention. Both backends now agree.

## Test

The test fake previously returned a dot product as `_distance`; it now models real L2 distance. New `test_lancedb_scores_are_similarity_higher_is_better` asserts converted scores are monotonic-with-closeness, bounded, exact-match `= 1.0`. Vector/embedding/llm suites pass (42).

Closes #126


tteon:

PR #236, opened 2026-06-09
fix(session): make ask_stream safe inside a running event loop

### LINKED ISSUE #125: Session.ask_stream creates a new event loop and fails when called inside a running loop (Jupyter/FastAPI)
## Required Information

**Describe the Bug:**

`Session.ask_stream` creates a fresh event loop with `asyncio.new_event_loop()` and drives the async iterator with `loop.run_until_complete(...)` directly (`seocho/session.py:836-864`), bypassing the worker-thread `_run_sync` path the rest of the Session uses. When `ask_stream` is called from a thread that already has a running event loop, such as a Jupyter cell or an async web handler (FastAPI), `run_until_complete` raises `RuntimeError: this event loop is already running` / `cannot be called from a running event loop`. The new loop is also never registered with `asyncio.set_event_loop`, and a brand-new loop is created on every call rather than reused.

**Steps to Reproduce:**

1. From inside a running event loop (a Jupyter notebook cell, or an `async def` FastAPI endpoint), call `session.ask_stream(question)` and iterate it.
2. A `RuntimeError` about a running event loop is raised.

**Expected Behavior:**

`ask_stream` works whether or not an event loop is already running, the same way `_run_sync` handles the synchronous `ask` path (run on a dedicated worker thread), or an explicit async streaming API is exposed for already-async callers.

**Observed Behavior:**

`ask_stream` raises `RuntimeError` whenever it is called from a thread with an active event loop.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

The non-streaming `ask` path already solves this with `_run_sync` (a worker thread that owns its own loop). Routing `ask_stream` through the same mechanism, or providing an `async` streaming method, would make it safe in notebooks and async servers.

Suggested labels: bug

1wos:
## Problem

`Session.ask_stream` drives its async iterator with `asyncio.new_event_loop()` + `loop.run_until_complete(...)`. When called from a thread that already has a running loop — a Jupyter cell or an async web handler (FastAPI) — this raises `RuntimeError: this event loop is already running`. The loop is also never registered with `set_event_loop` and is rebuilt on every call.

## Fix

Route streaming through the same worker-thread mechanism the synchronous `ask` path already uses (`_run_sync`). A new `_stream_async_in_thread` helper runs the async generator on a daemon thread that owns its own loop and forwards items through a `queue.Queue`, so:

- `ask_stream` is safe from any context (running loop or not),
- chunks still stream in real time (no buffering of the full answer),
- iterator cleanup (`aclose`) runs in the worker before the loop closes,
- exceptions propagate to the consumer after any partial output.

## Tests

`tests/seocho/test_stream_async_in_thread.py` pins both branches:
- streaming from a plain synchronous context,
- streaming from **inside** a running loop (the regression — `asyncio.run(_outer())`),
- exception propagation after partial output.

All 3 pass. The pre-existing `test_session_agent.py` DB-dependent failures are unrelated (they fail on `main` too).

Closes #125.


tteon:

PR #237, opened 2026-06-09
fix(chunk): track char offsets from the splitter so overlapped chunks keep provenance

### LINKED ISSUE #124: Chunk char offsets collapse to (-1,-1) for overlapped chunks, breaking source attribution
## Required Information

**Describe the Bug:**

`_locate_bodies` resolves each chunk body to a `(char_start, char_end)` span in the source with `content.find(body, max(search_start - 512, 0))` (`seocho/index/chunk.py:193`). For overlapped chunks this fails: the 200-char overlap prefix that gets prepended sits before the 512-char lookback window once a body is longer than about 512 characters, and the stripped and re-joined body no longer matches the source text verbatim, so `find` returns -1. The chunk then gets `(char_start, char_end) = (-1, -1)` (`:191`), which destroys the source attribution (provenance) that the chunk record is meant to carry.

**Steps to Reproduce:**

1. Chunk a document long enough to produce overlapped, multi-paragraph chunks (body length over ~512 chars with the default 200-char overlap).
2. Inspect the resulting chunk records.
3. The overlapped chunks have `char_start == -1` and `char_end == -1`.

**Expected Behavior:**

Every chunk resolves to its correct character span in the source, including overlapped chunks.

**Observed Behavior:**

Overlapped chunks collapse to `(-1, -1)`, losing the char offsets used for provenance and downstream highlighting.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

The locator could track offsets directly from the splitter (it already knows where each body was cut) instead of searching for the post-processed body string, which avoids the verbatim-match and lookback-window problems entirely.

Suggested labels: bug

1wos:
## Problem

`chunk()` resolved each body's source span *after* splitting, with `content.find(body, max(search_start - 512, 0))`. For overlapped, multi-paragraph chunks this returns `-1`, collapsing the chunk to `(char_start, char_end) = (-1, -1)` and destroying the source attribution the `Chunk` record exists to carry. Two effects combine:

1. the 200-char overlap prefix drifts out of the 512-char lookback window once a body exceeds ~512 chars;
2. paragraphs are `.strip()`-ed before being re-joined, so a body whose source had surrounding whitespace no longer matches verbatim and `find` returns `-1`.

## Fix

Track offsets during splitting instead of searching for the post-processed body (the approach suggested in the issue):

- `_paragraph_spans` recovers each paragraph's real `(start, end)` in the source, accounting for the leading-whitespace strip.
- `_split_with_offsets` carries those positions through the *same* chunking algorithm.

Body strings are byte-for-byte unchanged, so the `chunk_text` back-compat shim is unaffected. The span covers the chunk's own (non-overlap) content — first introduced paragraph to last paragraph — so spans tile the document without gaps; the borrowed overlap prefix stays attributed to the previous chunk. Empty input keeps the documented `(-1, -1)` sentinel.

## Tests

New `test_overlapped_long_chunks_keep_valid_offsets` reproduces the exact failing regime (bodies >512 chars, default 200 overlap, whitespace-padded paragraphs). It fails on `main` (every overlapped chunk is `-1`) and passes with this change; the source slice still resolves each chunk's own tail. All existing chunk tests pass unchanged.

Closes #124.


tteon:

PR #238, opened 2026-06-09
fix(packaging): add PyYAML to core dependencies so the public API imports

### LINKED ISSUE #137: pip install seocho then from seocho import Ontology fails with ModuleNotFoundError: yaml (PyYAML is not a core dependency)

1wos:
## Problem

`seocho/ontology.py` imports `yaml` at module top level, and `store/graph.py` re-imports `Ontology` eagerly. PyYAML was declared only in the `local`/`ci`/`dev` extras, so after a plain `pip install seocho`:

```
$ python -c "from seocho import Ontology"
ModuleNotFoundError: No module named 'yaml'
```

`import seocho` itself succeeded (lazy `__getattr__`), so the break only surfaced once a user touched the documented public API — `Ontology`, `OpenAIBackend`, `create_llm_backend`, `FAISSVectorStore`, the Neo4j/Ladybug stores. `seocho[ontology]` and `seocho[embedded]` also omit it.

## Fix

Move `pyyaml` into the core `[project] dependencies`.

## Tests

`tests/seocho/test_core_dependencies.py` parses `pyproject.toml` and asserts `pyyaml` is a core dependency (fails on `main`), plus pins the eager top-level `import yaml` that makes it mandatory.

Closes #137.


tteon:
Closing: already fixed on main — `"pyyaml"` is in the core `[project] dependencies` block of `pyproject.toml`. Thanks for catching it; it landed via another route.

PR #239, opened 2026-06-09
fix(file_reader): handle ragged CSV rows instead of crashing on None content

### LINKED ISSUE #140: CSV ingestion crashes with AttributeError when a data row is shorter than the header (None.strip())
## Required Information

**Describe the Bug:**

`read_csv_file` uses `csv.DictReader` (`seocho/index/file_reader.py:199`) and sets `content = row["content"]` (`:202`) when a `content` column exists. `DictReader` fills missing fields with `None` for any data row shorter than the header, so `content` becomes `None`, and `index_file` later calls `content.strip()` (`:424`), raising `AttributeError`. The reader call is wrapped in try/except, but the `index()` loop is not, so this crashes `index_file` rather than returning a clean `failed` result.

**Steps to Reproduce:**

1. Ingest a CSV whose header includes a `content` column but at least one data row has fewer fields than the header.
2. `AttributeError: 'NoneType' object has no attribute 'strip'` is raised and crashes ingestion.

**Expected Behavior:**

Ragged or short rows are handled (skipped or coerced to ""), and a malformed file produces a clean `failed` status rather than crashing the run.

**Observed Behavior:**

`None.strip()` raises an unhandled `AttributeError`.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`csv.DictReader` fills a row shorter than the header with `None` for the missing fields. When a `content` column exists, `content` became `None`, and `index_file`'s later `content.strip()` raised `AttributeError: 'NoneType' object has no attribute 'strip'`. The reader call is wrapped in try/except but the `index()` loop is not, so a single malformed row crashed the whole ingestion run instead of yielding a clean `failed` status.

## Fix

- Read with `restval=""` so short rows fill missing fields with `""` instead of `None`.
- Coerce a missing/None `content` to `""`.
- Drop the `None` restkey DictReader uses to collect overflow from rows *longer* than the header, keeping metadata string-keyed and JSON-serializable.

## Tests

- `test_csv_short_row_does_not_crash` — a row shorter than the header yields `content == ""` (a real string, so `.strip()` is safe). Fails on `main` (`content is None`).
- `test_csv_long_row_drops_overflow_restkey` — an over-long row drops the `None` key and stays serializable.

Closes #140.


tteon:

PR #240, opened 2026-06-09
fix(answering): report financial delta in the year order stated in the question

### LINKED ISSUE #131: Financial delta reports the wrong direction for questions phrased from a later year to an earlier year
## Required Information

**Describe the Bug:**

`financial_metric_delta` computes `end_row["value"] - start_row["value"]` where start and end come from `_ordered_years`, which orders years chronologically regardless of how the question is phrased (`seocho/query/answering.py:125`, with `_ordered_years` around `:92`/`:106`). A question like "change in revenue from 2023 to 2021" is answered as 2021 to 2023, so the reported increased/decreased direction and the sign are reversed relative to the question.

**Steps to Reproduce:**

1. Ask a delta question with the later year first, e.g. "how did revenue change from 2023 to 2021".
2. The answer's direction and sign are inverted versus the question.

**Expected Behavior:**

The delta direction follows the order stated in the question.

**Observed Behavior:**

Direction always follows ascending years, reversing descending-phrased questions.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`financial_metric_delta` computed `end_row["value"] - start_row["value"]` with start/end taken from `_ordered_years`, which sorts years chronologically regardless of phrasing. A question like "how did revenue change from 2023 to 2021" was answered as 2021 → 2023, so the reported direction (increased/decreased) and the sign were reversed versus the question.

## Fix

Add `_dedupe_years` (dedupe preserving stated order) and have the delta branch take its endpoints from the question-stated order. Fall back to chronological order only when the question did not name the years. `_ordered_years` now builds on `_dedupe_years`, so every chronological caller is byte-for-byte unchanged.

## Tests

`test_financial_delta_direction.py`:
- descending phrasing ("from 2023 to 2021") → **decreased**, reported `from 2023 to 2021` (fails on `main`, which says "increased … from 2021 to 2023");
- ascending phrasing → increased;
- no years named → chronological fallback.

Closes #131.


tteon:

PR #241, opened 2026-06-09
fix(ontology): score a property type mismatch as 0, not 0.5 partial credit

### LINKED ISSUE #129: score_extraction gives 0.5 partial credit on property type mismatch, inflating scores and defeating the quality-retry gate
## Required Information

**Describe the Bug:**

`score_extraction` type-correctness adds `+0.5` partial credit when a property value's type does not match the schema (`seocho/ontology.py:1766`, commented "partial credit for other types"). Because STRING is the default type for most properties, genuinely malformed extractions still score high (around 0.8+). The indexing quality-retry gate compares against this inflated `overall`, so it rarely triggers and low-quality extractions pass.

**Steps to Reproduce:**

1. Produce an extraction with several wrong-typed property values.
2. Call `score_extraction(...)`.
3. `overall` stays above a typical `quality_threshold`, so no quality retry fires during indexing.

**Expected Behavior:**

A type mismatch scores 0 (or the score does not mask malformed extractions enough to disable the retry gate).

**Observed Behavior:**

0.5 partial credit inflates the score and suppresses the quality retry.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`score_extraction`'s type-correctness loop added `+0.5` partial credit whenever a property value's type did not match the schema ("partial credit for other types"). Since STRING is the default type for most properties, genuinely malformed extractions still scored ~0.8+ overall. The indexing quality-retry gate compares against this inflated `overall`, so it rarely triggered and low-quality extractions passed.

## Fix

- A genuine type mismatch now scores `0`.
- The old `+0.5` `else` branch was also the only handling for `LIST` and `POINT` property types, so add explicit checks (`list`/`tuple` for LIST, `dict`/`list`/`tuple` for POINT). A correctly-typed LIST/POINT keeps full credit; a mistyped one scores 0.

## Tests (`TestConfidenceScoring`)

- `test_type_mismatch_scores_zero_not_partial` — INTEGER property given a string → `type_correctness == 0.5` for a node with one good + one bad property (was `0.75` on main).
- `test_all_wrong_types_zero_type_correctness` — all properties mis-typed → `0.0` (was `0.5`, the gate-defeating case).
- `test_list_property_correct_type_is_credited` — correct LIST → `1.0`, mistyped → `0.0`.

Both regression tests fail on `main`; the full `test_ontology.py` suite stays green.

Closes #129.


tteon:

PR #242, opened 2026-06-09
fix(ontology): detect relationship cardinality conflicts on merge

### LINKED ISSUE #130: Ontology.merge silently keeps the left relationship on source/target or cardinality conflict instead of reporting it
## Required Information

**Describe the Bug:**

When two modules define the same relationship with different source/target, `_merge_rel_def` records a conflict and then `return left` early (`seocho/ontology.py:2152`+), so even in strict mode the later raise-on-conflicts is bypassed for that relationship. Relationship cardinality differences (for example ONE_TO_MANY vs MANY_TO_MANY) are never compared or recorded at all, so a merge silently keeps the left cardinality with no signal.

**Steps to Reproduce:**

1. Merge two ontologies that define the same relationship with different cardinality (and/or source/target).
2. Inspect the merged relationship and any reported conflicts.
3. The merged relationship keeps the left definition; cardinality conflict is not recorded, and strict mode does not raise on the source/target path.

**Expected Behavior:**

Conflicts, including cardinality, are recorded, and strict mode raises.

**Observed Behavior:**

Left wins silently; cardinality conflicts are never detected.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`_merge_rel_def` compared relationship source/target but **never compared cardinality**. Merging two ontologies that defined the same relationship with different cardinality (e.g. `ONE_TO_MANY` vs `MANY_TO_MANY`) silently kept the left side with no conflict recorded, so even `strategy="strict"` did not raise. It also returned early on a source/target conflict in strict mode, which could mask a simultaneous cardinality conflict.

## Fix

- Compare cardinality (case/whitespace-normalized) and append a conflict when it differs.
- Drop the early `return left` on a source/target conflict; `merge()` raises on the accumulated conflict list afterward, so falling through records *all* conflicts in one pass.

## Tests (`test_ontology_merge_conflicts.py`)

- `test_cardinality_conflict_raises_in_strict_mode` — fails on `main` (silent), passes here.
- `test_cardinality_conflict_is_silent_outside_strict` — union keeps left cardinality, no raise (behavior preserved).
- `test_source_target_conflict_still_raises_in_strict_mode` — regression guard.
- `test_matching_relationships_merge_without_conflict`.

Closes #130.


tteon:

PR #243, opened 2026-06-09
fix(index): scope content dedup to a single batch, not the pipeline lifetime

### LINKED ISSUE #133: IndexingPipeline._seen_hashes is shared across documents, so identical text in separate add() calls is silently deduplicated
## Required Information

**Describe the Bug:**

`_seen_hashes` is instance state on a long-lived pipeline (`seocho/index/pipeline.py:244`, used around `:669`). It persists across separate `add()` calls, so a second document whose normalized text hashes to one already seen returns `deduplicated=True` and writes nothing, even though the caller meant to index a distinct document. The state is also per-process only, so behavior depends on process lifetime.

**Steps to Reproduce:**

1. On one client, call `add(text)`.
2. Call `add(text)` again with the same normalized text.
3. The second call returns `deduplicated=True` and writes no graph.

**Expected Behavior:**

Dedup is scoped to a single ingest, or is opt-in and documented, rather than silently shared across independent `add()` calls.

**Observed Behavior:**

Cross-call dedup drops legitimately re-submitted documents.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`_seen_hashes` was instance state on a long-lived `IndexingPipeline`, so it persisted across independent `index()`/`add()` calls. A second document whose normalized text hashed to one already seen returned `deduplicated=True` and wrote nothing — even though the caller meant to index a distinct document. Behavior also depended on process lifetime.

## Fix

Scope dedup to a single batch ingest:
- `index_batch` owns a fresh `set()` per call and threads it into `index()` via a new `_seen_hashes` parameter.
- standalone `index()` calls pass `None` and are never deduplicated against earlier, independent calls.
- `reindex` no longer clears a shared cache (there isn't one).
- `enable_dedup` still gates within-batch dedup.

## Tests (`test_dedup_scope.py`)

- `test_no_process_lifetime_dedup_state` — the instance set is gone.
- `test_duplicate_within_batch_set_is_skipped_early` — a within-batch duplicate short-circuits without touching graph/llm.
- `test_each_batch_gets_a_fresh_dedup_set` — two `index_batch` calls receive distinct sets (same set within a batch, fresh set across batches), proving no cross-call leak.
- existing `test_indexing.py` stays green.

Closes #133.


tteon:

PR #244, opened 2026-06-09
fix(tracing): configure Opik via constructor, not by mutating os.environ

### LINKED ISSUE #141: OpikBackend mutates process environment variables (including OPIK_API_KEY) at init
## Required Information

**Describe the Bug:**

When the Opik tracing backend initializes, it writes its configuration into `os.environ`, including the API key (`seocho/tracing.py:125-130`):

```python
os.environ["OPIK_URL_OVERRIDE"] = self._url
os.environ["OPIK_WORKSPACE"] = self._workspace
os.environ["OPIK_PROJECT_NAME"] = self._project
os.environ["OPIK_API_KEY"] = self._api_key
```

A library mutating the host process environment is a surprising side effect. It overwrites any `OPIK_*` the caller (or another component / session in the same process) had set, and the new values are inherited by child processes. Writing `OPIK_API_KEY` into `os.environ` also widens the key's exposure (child processes, anything that dumps the environment, crash logs). In a process where more than one Opik configuration is used, or where the host application also talks to Opik, this silently changes global state behind the caller's back.

**Steps to Reproduce:**

1. Set `OPIK_PROJECT_NAME=A` in the environment.
2. Initialize a SEOCHO Opik tracing backend configured for project `B`.
3. `os.environ["OPIK_PROJECT_NAME"]` is now `B` for the rest of the process and any child processes; the original `A` is lost.

**Expected Behavior:**

The backend reads configuration from env/args but does not write back into `os.environ`. The Opik client is configured through its constructor (`project_name` / `workspace` / `api_key` / `host`) instead.

**Observed Behavior:**

Init overwrites `OPIK_URL_OVERRIDE`, `OPIK_WORKSPACE`, `OPIK_PROJECT_NAME`, and `OPIK_API_KEY` in the process environment.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

Passing the values directly to the `opik.Opik(...)` constructor instead of writing them to `os.environ` removes the side effect and also avoids cross-configuration contention when several Opik sessions share one process or machine.

Suggested labels: bug

1wos:
## Problem

`OpikBackend.__init__` wrote `OPIK_URL_OVERRIDE`, `OPIK_WORKSPACE`, `OPIK_PROJECT_NAME`, and `OPIK_API_KEY` into `os.environ`. A library mutating the host process environment is a surprising side effect: it clobbered any `OPIK_*` the caller (or another component/session in the same process) had set, the values leaked into child processes, and writing `OPIK_API_KEY` widened the key's exposure (child processes, env dumps, crash logs).

## Fix

Configure the client through its constructor — `opik.Opik(project_name=..., workspace=..., host=..., api_key=...)` — and stop writing to `os.environ`.

## Tests

- `test_init_does_not_mutate_os_environ` — pre-set `OPIK_*` sentinels survive init unchanged (fails on `main`).
- `test_config_passed_through_constructor` — `project_name`/`workspace`/`host`/`api_key` reach the `Opik(...)` constructor (fails on `main`, which passed only `project_name`).
- existing opik regression + tracing tests stay green.

Closes #141.


tteon:
The code this PR fixes no longer exists — the Opik tracing backend was removed in #518 (ADR-0166). Closing as obsolete rather than rejected.

Thank you for the fix.

PR #245, opened 2026-06-09
fix(vector): validate FAISS embedding dimension with a clear error

### LINKED ISSUE #121: FAISSVectorStore does not validate embedding dimension, crashing on non-1536 embedding models
## Required Information

**Describe the Bug:**

`FAISSVectorStore` builds its index as `faiss.IndexFlatIP(dimension)` with a fixed default dimension (1536, matching `text-embedding-3-small`) and never checks that incoming vectors match it. `add()` passes vectors straight to `self._index.add(vectors)` (`seocho/store/vector.py:101` and the `add` body around `:120`) with no comparison between `self._dimension` and the embedding width returned by the backend. Configuring any model with a different embedding size, for example `text-embedding-3-large` (3072) or many vLLM/Qwen presets, makes the first `add()` raise an opaque FAISS assertion in the C layer instead of a clear, actionable error.

**Steps to Reproduce:**

1. Configure embeddings with a non-1536 model (e.g. `text-embedding-3-large`, 3072 dims) while `FAISSVectorStore` keeps its default dimension.
2. Index any document so a vector reaches `add()`.
3. FAISS raises an assertion error from native code; the message does not mention the dimension mismatch.

**Expected Behavior:**

The dimension is derived from the first embedding (or validated against `self._dimension`), and a mismatch raises a clear Python error naming the expected and actual dimensions.

**Observed Behavior:**

An opaque FAISS native assertion on the first `add()` whenever the model's embedding width differs from the hardcoded default.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

A caller-side workaround is to construct `FAISSVectorStore(dimension=...)` matching the configured model. Deriving the dimension lazily from the first vector would remove the footgun entirely.

Suggested labels: bug

1wos:
## Problem

`FAISSVectorStore` built `faiss.IndexFlatIP` with a fixed default dimension (1536) and never checked incoming vector width. Configuring any model with a different embedding size (e.g. `text-embedding-3-large` at 3072, many vLLM/Qwen presets) made the first `add()` raise an opaque assertion from faiss's native layer that never mentioned the dimension mismatch.

## Fix

`_ensure_dimension`:
- adopts the first embedding's width when the index is still empty (so any model just works), and
- raises a clear `ValueError` naming the expected and actual dimensions when a later vector disagrees.

Applied to both `add()` and `add_batch()`.

## Tests (`test_faiss_vector_store.py`, `importorskip("faiss")`)

- `test_first_embedding_dimension_is_adopted` / `test_add_batch_adopts_first_dimension` — width-8/16 vectors are adopted over the 1536 default (fail on `main` with a native faiss assertion).
- `test_dimension_mismatch_raises_clear_error` — a later differing width raises `ValueError` (fails on `main`, which raised a native `AssertionError`).

Closes #121.


tteon:

PR #246, opened 2026-06-09
fix(vector): keep FAISS search returning limit live results after deletes

### LINKED ISSUE #122: FAISSVectorStore.delete only tombstones the doc and leaves the vector in the index, polluting search
## Required Information

**Describe the Bug:**

`FAISSVectorStore.delete()` marks a document as deleted by setting `self._docs[idx]["_deleted"] = True` (`seocho/store/vector.py:177-187`) but never removes the vector from the FAISS index. The tombstoned vector still participates in `search()`, and results are filtered after the fact by the `_deleted` flag. Two consequences follow: deleted vectors consume top-k slots, so a `search(limit=k)` can return fewer than `k` live results even when more live vectors exist; and the index never reclaims space, so `ntotal` keeps growing across deletes.

**Steps to Reproduce:**

1. Add several documents, then `delete()` some of them.
2. `search(query, limit=k)` where at least `k` live documents remain.
3. Fewer than `k` live results come back, because deleted vectors that score highly still occupy slots before the post-filter drops them.

**Expected Behavior:**

Deleted vectors are removed from the index (or the search over-fetches and then filters so it still returns `k` live results), and the index size reflects live documents.

**Observed Behavior:**

Deleted vectors remain in the index, are scored during search, and shrink the effective number of live results returned.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

`IndexFlatIP` does not support in-place removal, so a fix likely means rebuilding the index on delete, switching to an `IndexIDMap`/`remove_ids`-capable index, or over-fetching by the deleted count before filtering.

Suggested labels: bug

1wos:
## Problem

`FAISSVectorStore.delete()` tombstones a doc (`_deleted = True`) but cannot remove its vector from an `IndexFlatIP`, so the deleted vector still participated in `search()` and was filtered only after the fact. A tombstoned vector that scored highly consumed a top-k slot, so `search(limit=k)` could return fewer than `k` live results even when more live vectors existed.

## Fix

Track the tombstone count and have `search()` over-fetch by it (bounded by `ntotal`), stopping once `limit` live results are collected. `count()` already reports live docs.

## Tests (`test_faiss_delete.py`, `importorskip("faiss")`)

- `test_delete_still_returns_full_limit_of_live_results` — the top-ranked doc is deleted; a `limit=2` search over 3 remaining live docs returns 2 (fails on `main`, which returned 1).
- `test_deleted_doc_not_in_results`, `test_count_reflects_live_docs_after_delete`.

Closes #122.


tteon:

PR #247, opened 2026-06-09
fix(artifacts): guard approve against reviving a deprecated artifact

### LINKED ISSUE #138: approve_semantic_artifact has no status guard, so a deprecated artifact can be silently revived to approved
## Required Information

**Describe the Bug:**

`approve_semantic_artifact` (`extraction/semantic_artifact_store.py:188`) sets `payload["status"] = "approved"` unconditionally and resets `deprecated_by` / `deprecation_note` / `deprecated_at` to None, with no check on the current status. A previously `deprecated` artifact can therefore be re-approved with no error, and is then served to the runtime through `resolve_approved_artifact_payload`. The lifecycle has no guarded transitions.

**Steps to Reproduce:**

1. Create and approve an artifact, then deprecate it.
2. Call `approve_semantic_artifact` on the same id again.
3. It returns to `approved` with the deprecation fields cleared, and is served as the active artifact.

**Expected Behavior:**

`approve` rejects (or requires an explicit override) when the artifact is deprecated; status transitions are validated.

**Observed Behavior:**

deprecated to approved succeeds silently with no guard.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`approve_semantic_artifact` set `status="approved"` unconditionally and reset `deprecated_by`/`deprecation_note`/`deprecated_at` to `None` with no check on the current status. A previously **deprecated** artifact could be re-approved with no error and then served to the runtime via `resolve_approved_artifact_payload` — the lifecycle had no guarded transitions.

## Fix

Reject approval when the artifact is already `deprecated` (end-of-life), mirroring the existing `deprecate` guard (`must be approved before deprecation`). `draft → approved` and idempotent re-approval of an `approved` artifact stay permitted.

## Tests (`test_semantic_artifact_lifecycle.py`, offline via `base_dir=tmp_path`)

- `test_deprecated_artifact_cannot_be_reapproved` — raises `ValueError` (fails on `main`, which revived it).
- `test_draft_can_be_approved_then_deprecated`, `test_reapproving_an_approved_artifact_is_allowed`.

Closes #138.


tteon:

PR #249, opened 2026-06-09
fix(artifacts): write semantic artifacts atomically and skip corrupt files on list

### LINKED ISSUE #139: Semantic artifact writes are not atomic; a crash mid-write corrupts the artifact and breaks the list endpoint
## Required Information

**Describe the Bug:**

`create`, `approve`, and `deprecate` write artifacts with `open(path, "w")` followed by `json.dump(...)` (`extraction/semantic_artifact_store.py:58, 197, 227`), with no temp-file-then-rename. A crash or concurrent write during `json.dump` leaves a truncated file, and every later `json.load` on that artifact (get / approve / deprecate) raises, making it unrecoverable. `list_semantic_artifacts` (`:152`) also `json.load`s every `*.json` with no try/except, so a single corrupt or partially written file makes the entire list endpoint raise (500).

**Steps to Reproduce:**

1. Interrupt the process while an artifact is being written (or write the same artifact concurrently).
2. The artifact JSON is left truncated.
3. Reads of that artifact raise, and `list_semantic_artifacts` fails for the whole store.

**Expected Behavior:**

Writes are atomic (write to a temp file, then `os.replace`), and the list path tolerates a corrupt file instead of failing entirely.

**Observed Behavior:**

Truncate-in-place writes; one corrupt file breaks both that artifact and listing.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`create`/`approve`/`deprecate` wrote with `open(path, "w")` + `json.dump` and no temp-file-then-rename. A crash or concurrent write during `json.dump` left a truncated file, and every later `json.load` on it (get/approve/deprecate) raised — making the artifact unrecoverable. `list_semantic_artifacts` `json.load`'d every `*.json` with no guard, so a single corrupt/partially-written file made the whole list endpoint raise (500).

## Fix

- `_atomic_write_json`: serialize to a temp file in the same directory, `fsync`, then `os.replace` into place. All four write sites route through it.
- `list_semantic_artifacts` skips (and logs) an unreadable artifact instead of raising for the entire store.

## Tests (`test_semantic_artifact_durability.py`)

- `test_corrupt_artifact_does_not_break_listing` — a truncated `*.json` is skipped; valid artifacts still listed (on `main`, `list_semantic_artifacts` raises `JSONDecodeError`).
- `test_save_leaves_no_temp_files`, `test_atomic_write_roundtrip_and_overwrite`.

Closes #139.


tteon:

PR #252, opened 2026-06-09
fix(extraction): select prompt from a per-call category, not shared instance state

### LINKED ISSUE #134: ExtractionStrategy.category is mutated on a shared strategy instance per call, racing under concurrent add()
## Required Information

**Describe the Bug:**

`_LocalEngine` reuses one `ExtractionStrategy` and sets `self._extraction.category = category` per call; `ExtractionStrategy.render` then reads `self.category` to choose the prompt (`seocho/query/strategy.py:391`, `:406`). Two concurrent `add()` calls on the same Session/engine race on the shared `category`, so one call can render with the other call's category and select the wrong prompt.

**Steps to Reproduce:**

1. Issue two concurrent `add()` calls with different `category` values on one client.
2. Prompt selection can be corrupted by the interleaving (a call renders with the other's category).

**Expected Behavior:**

Category is passed per call without shared mutable state, so concurrent calls do not interfere.

**Observed Behavior:**

Shared `category` mutation races across concurrent calls.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`_LocalEngine` reused one `ExtractionStrategy` and set `self._extraction.category = category` per call; `ExtractionStrategy.render` then read `self.category` to choose the prompt. Two concurrent `add()` calls on one Session/engine raced on the shared `category`, so one call could render with the other's category and select the wrong prompt. The same mutate-then-render pattern existed in the indexing extraction engine, the pipeline retry path, and the `extract` tool.

## Fix

`render()` reads an optional per-call `category` kwarg (`kwargs.get("category") or self.category`) and uses it for both the preset auto-select and the system-prompt cache key. Every call site now passes `category=category` instead of mutating the shared instance:

- `local_engine.py` `extract`
- `index/extraction_engine.py`
- `index/pipeline.py` retry render
- `tools.py` extract tool

No mutable cross-call state remains on the prompt path.

## Tests (`test_extraction_category_per_call.py`)

- `test_render_uses_per_call_category` — `render(category="Financials")` vs `"Legal"` select the matching presets (fails on `main`, which ignores the kwarg and uses `self.category`).
- `test_render_does_not_mutate_instance_category` — `self.category` is untouched.
- `test_concurrent_renders_do_not_cross_talk` — 16 threads × 50 renders with 4 categories, no cross-talk.

Closes #134.


tteon:

PR #256, opened 2026-06-09
feat(ontology): declare and enforce enum/range property constraints in SHACL

### LINKED ISSUE #128: enum and range constraints are declared but never enforced during indexing (validate_with_shacl ignores them)
## Required Information

**Describe the Bug:**

`to_shacl()` emits only `datatype` / `minCount` / `maxCount` / unique shapes (`seocho/ontology.py:962`+); it never emits `sh:in` (enum) or min/max (range). `validate_with_shacl`, the path `IndexingPipeline` calls, checks datatype and cardinality only. The `rules.py` engine defines `enum` and `range` rule kinds and `_validate_value` can enforce them, but the indexing validation path never consults inferred rules, so enum/range constraints declared on an ontology are silently unenforced when writing the graph.

**Steps to Reproduce:**

1. Declare a property with an enum (allowed value set) on the ontology.
2. Index text that yields an out-of-enum value.
3. The value is written with no validation error.

**Expected Behavior:**

enum and range constraints are enforced (or at least surfaced) during indexing validation.

**Observed Behavior:**

Only datatype and cardinality are checked; enum/range are ignored on the indexing path.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`to_shacl()` emitted only `datatype` / `minCount` / `maxCount` / unique shapes, and `validate_with_shacl` (the path `IndexingPipeline` calls) checked datatype and cardinality only. So an `enum` (allowed value set) or numeric range constraint on a property was **silently unenforced** when writing the graph. The `rules.py` engine can enforce enum/range, but only on the opt-in inferred-rules path (`enable_rule_constraints`), never on the always-on indexing validation.

## Fix

- Add `enum: Optional[List]` and `value_range: Optional[Tuple[float, float]]` to `Property`.
- Persist them through `to_dict` / `_from_dict` (YAML/JSON roundtrip).
- Emit `sh:in` / `sh:minInclusive` / `sh:maxInclusive` from `to_shacl`.
- Enforce both in `validate_with_shacl`: an out-of-set or out-of-range value now produces a validation error.

## Tests (`test_shacl_enum_range.py`)

- `to_shacl` emits `in` / `minInclusive` / `maxInclusive`.
- enum/range violations are reported; valid members pass.
- enum/range survive a `to_dict` → `from_dict` roundtrip.

The full `test_ontology.py` suite stays green; the enforcement tests fail on `main`.

Closes #128.


tteon:

PR #259, opened 2026-06-09
fix(graph): release Neo4jGraphStore driver via context manager and __del__

### LINKED ISSUE #135: Neo4jGraphStore opens a driver in __init__ that is never closed by the engine/session, leaking connection pools
## Required Information

**Describe the Bug:**

`Neo4jGraphStore` opens the driver in `__init__` (`seocho/store/graph.py:323`) and exposes `close()` (`:290`), but there is no `__del__`, no context-manager support, and neither `_LocalEngine` nor `Session.close()` call `graph_store.close()`. Each store therefore holds an open Neo4j driver connection pool for the process lifetime, so creating stores per request or per workspace leaks pools.

**Steps to Reproduce:**

1. Create several `Seocho` clients over the life of a process, each with its own `Neo4jGraphStore`, without manually calling `close()`.
2. Driver connection pools accumulate and are never released.

**Expected Behavior:**

The driver is closed when the store, engine, or session is closed (context manager or a cascading `close()`).

**Observed Behavior:**

Driver is opened in `__init__` and never auto-closed; pools leak.

**Environment Details:**

- OS: macOS
- Python: 3.12

Suggested labels: bug

1wos:
## Problem

`Neo4jGraphStore` opened the driver in `__init__` and exposed `close()`, but had no `__del__`, no context-manager support, and nothing (`_LocalEngine` / `Session.close()`) cascaded a `close()`. A store created per request or per workspace held its Neo4j driver connection pool for the whole process lifetime, leaking pools.

## Fix

- `__enter__`/`__exit__` → `with Neo4jGraphStore(...) as store:` releases the pool deterministically.
- a guarded `__del__` as a best-effort safety net for stores that go out of scope without an explicit `close()` (guarded because `__del__` can run during interpreter shutdown).
- `close()` is now idempotent so the explicit/`__exit__`/`__del__` paths cannot double-close the driver.

This is intentionally non-invasive: it does not auto-close a user-supplied store from `Session.close()` (the caller may still be using it), but gives deterministic (`with`) and safety-net (`__del__`) release.

## Tests (`test_neo4j_store_close.py`, fake `neo4j` module)

- context manager closes the driver exactly once (fails on `main`);
- `close()` is idempotent (fails on `main`, which closed twice);
- `__del__` closes as a safety net, and does not double-close after an explicit `close()`.

Closes #135.


tteon:

PR #260, opened 2026-06-09
fix(tracing): construct Opik client with batching=False to keep burst traces

### LINKED ISSUE #119: OpikBackend loses trace payloads under burst because opik.Opik() is created with default batching
## Required Information

**Describe the Bug:**

`OpikBackend` constructs its client as `opik.Opik(project_name=self._project)` (`seocho/tracing.py:133`) without `batching=False`. Opik's default asynchronous batching coalesces spans and flushes them in the background, and under a burst of short-lived traces, each created and immediately `end()`ed, which is exactly how `log_extraction` and `log_query` emit, the payloads are dropped before they flush. The traces then appear with a missing name, tags, or body, or do not appear at all. In our runs, logging N traces in quick succession persisted roughly 1 of N with the default client, versus N of N once the client was created with `batching=False`.

**Steps to Reproduce:**

1. Set `SEOCHO_TRACE_BACKEND=opik` with hosted Opik credentials.
2. Index several documents back to back, or run any loop that emits many `sdk.extraction` traces in quick succession.
3. Open the Opik project: a subset of the expected traces are missing, or arrive without name, tags, or payload.

**Expected Behavior:**

Every emitted trace is persisted. The backend either constructs the client with `batching=False`, or flushes after each span, or documents that callers must disable batching for burst workloads.

**Observed Behavior:**

Under burst emission most traces are silently lost, and only a fraction reach the project.

**Environment Details:**

- OS: macOS
- Python: 3.12
- Opik: cloud (`SEOCHO_TRACE_BACKEND=opik`, hosted)

---

## Optional Information

`flush()` exists on the backend (`seocho/tracing.py`, around line 166) but only runs from `disable_tracing()` and `close()`, not per span, so it does not help mid-burst. A caller-side workaround is to drive an `opik.Opik(batching=False)` client directly instead of relying on the SDK tracing path.

Suggested labels: bug

1wos:
## Problem

`OpikBackend` built its client as `opik.Opik(project_name=self._project)` with Opik's default async batching. Batching coalesces spans and flushes in the background, so a burst of short-lived traces — each created and immediately `end()`ed, exactly how `log_extraction`/`log_query` emit — was dropped before it flushed. In our runs only ~1 of N traces persisted with the default client, versus N of N with `batching=False`.

## Fix

Construct the client with `batching=False`.

## Tests (`test_opik_batching.py`, fake `opik` module)

Captures the `opik.Opik(...)` constructor kwargs and asserts `batching is False` (fails on `main`). Existing tracing/opik regression tests stay green.

> Note: touches the same `opik.Opik(...)` construction as #141 (PR #244); whichever merges first, the other is a trivial rebase.

Closes #119.


tteon:
Closing: fixed by other means. Main solved the dropped-burst-trace symptom by passing `end_time=` in the single `trace()` create call instead of create-then-`end()`, with a comment citing exactly this batching race (`src/seocho/tracing.py`). Merging `batching=False` on top would be a second mechanism for a symptom that no longer occurs.

PR #261, opened 2026-06-09
fix(index): delete a source's vectors on delete_source/reindex

### LINKED ISSUE #123: reindex and delete_source remove graph nodes but never delete their vectors, leaving orphans
## Required Information

**Describe the Bug:**

`reindex()` and `delete_source()` remove a source's graph nodes and clear the dedup hash, but never delete the source's vectors from the attached vector store. `delete_source()` calls `self.graph_store.delete_by_source(...)` only (`seocho/index/pipeline.py:1280`+), and `reindex()` calls `delete_source()` then discards `_seen_hashes` (`:1258`+); neither calls `vector_store.delete()`. After a reindex or a source delete, the previous version's vectors stay in the store and keep surfacing as top-k hits that point at graph nodes which no longer exist.

This is distinct from the already-filed write-path graph/vector inconsistency: this is the delete and reindex path leaking vectors.

**Steps to Reproduce:**

1. Index a source with a vector store attached.
2. `reindex(source_id)` (or `delete_source(source_id)`).
3. Run a vector search: stale vectors from the deleted/old version are returned and resolve to missing or stale graph nodes.

**Expected Behavior:**

`delete_source()` and `reindex()` also delete the source's vectors so the vector store and graph stay consistent.

**Observed Behavior:**

Orphan vectors persist after delete/reindex and produce stale retrieval hits.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

The vector records are keyed per source/chunk, so the delete path has the information it needs to remove them; it just never calls into the vector store.

Suggested labels: bug

1wos:
## Problem

`delete_source()` removed only graph nodes (`graph_store.delete_by_source`), and `reindex()` goes through `delete_source()` — so neither dropped the source's vectors from the attached vector store. After a delete or reindex the previous version's vectors stayed in the store and kept surfacing as top-k hits that resolve to missing/stale graph nodes.

## Fix

- Add `VectorStore.delete_by_source(source_id) -> int`: a no-op default (custom stores), overridden by:
  - **FAISS** — match on `metadata.source_id` or the `{source_id}_chunk_` id prefix;
  - **LanceDB** — match on the id prefix (metadata is stored as a JSON string, so it isn't directly queryable).
- `delete_source()` calls it and reports `vectors_deleted` in the summary. `reindex()` is fixed transitively.

## Tests (`test_delete_source_vectors.py`)

- pipeline `delete_source` calls `vector_store.delete_by_source` and surfaces `vectors_deleted` (fails on `main`); no-op when no vector store is attached.
- FAISS end-to-end (`importorskip`): `delete_by_source("srcA")` removes exactly srcA's 3 vectors, leaving srcB's — fails on `main` (no such method).

Closes #123.


tteon:

PR #263, opened 2026-06-09
fix(index): raise on extraction failure by default and raise reasoning timeouts

### LINKED ISSUE #118: Extraction silently falls back to generic Entity nodes and returns success; default 120s timeout is too short for reasoning models
## Required Information

**Describe the Bug:**

When the per-chunk extraction LLM call raises, including a timeout, the indexing pipeline does not surface the failure. It calls `_fallback_extract`, which produces generic `Entity` nodes that ignore the active ontology, sets `fallback_used=True`, and continues (`seocho/index/pipeline.py:708-718`, and the empty-graph branch at `:724-732`). `IndexingResult.ok` stays `True` and `add()` returns a normal `Memory`, so a degraded, ontology-free extraction is reported as success, with the only signal buried in `result.metadata["fallback_used"]`. `docs/BASELINE_INSTRUCTIONS.md` section 1.1 ("Silent-fallback discipline") states the opposite: "when an SDK or runtime path fails, raise an explicit exception. Never silently substitute a degraded path and return success," with degraded fallback gated behind an explicit flag. The extraction path has no such flag.

This is easy to hit because the default timeout is short. `OpenAICompatibleBackend`, and its `KimiBackend` subclass, default to `timeout: float = 120.0` (`seocho/store/llm.py:311`), but the shipped reasoning presets are slow: with `kimi/kimi-k2.5`, single-document extraction took 160 to 1450 seconds in our runs, so the 120s ceiling is exceeded routinely and silently trips the fallback above.

**Steps to Reproduce:**

```python
# SEOCHO_LLM=kimi/kimi-k2.5 (a reasoning model, default timeout 120s)
mem = client.add(long_financial_text, category="Financials")

# if the extraction call exceeds 120s it times out, the pipeline falls back to
# heuristic extraction, and add() returns success
mem.metadata["fallback_used"]   # True  (the only indication)
# the graph now holds generic Entity nodes instead of the ontology's labels
```

**Expected Behavior:**

A failed extraction raises, or is gated behind an explicit opt-in (for example `on_failure="degrade"`, per baseline 1.1), and `IndexingResult.ok` reflects the degradation. The default timeout for the reasoning-model presets is high enough that a normal extraction is not cut off.

**Observed Behavior:**

Extraction timeouts are absorbed into a heuristic fallback that returns success with ontology-free `Entity` nodes, and the default 120s timeout makes this the common case for reasoning models.

**Environment Details:**

- OS: macOS
- Python: 3.12

---

## Optional Information

The two halves reinforce each other: raising the default timeout for the reasoning presets removes most of the silent fallbacks, and honoring baseline 1.1 (raise, or require an explicit opt-in) makes the remaining ones visible. A caller-side workaround today is to construct the backend with a larger `timeout` and to check `mem.metadata["fallback_used"]` after every `add()`.

Suggested labels: bug

1wos:
## Problem

Two coupled defects:

1. **Silent-fallback discipline.** When the per-chunk extraction call raised (including a timeout), the pipeline called `_fallback_extract`, produced ontology-free `Entity` nodes, set `fallback_used=True`, and continued — with `IndexingResult.ok` still `True`. A degraded, ontology-free extraction was reported as success, with the only signal buried in `fallback_used`. `BASELINE_INSTRUCTIONS.md` §1.1 mandates the opposite: raise, with degradation gated behind an explicit flag.
2. **Timeout too short.** `OpenAICompatibleBackend`/`KimiBackend` default to `timeout=120.0`, but reasoning presets are slow (kimi-k2.5 single-document extraction measured at 160–1450s), so the ceiling was exceeded routinely and silently tripped the fallback.

## Fix

1. `IndexingPipeline(extraction_on_failure=...)` — `"raise"` (default) surfaces the failure; `"degrade"` opts into the heuristic fallback. `IndexingResult.ok` now returns `False` whenever a fallback was used, so degradation is visible to callers.
2. Add `ProviderSpec.default_timeout` (default 120s), set it to **900s** for the `kimi` and `grok` reasoning presets, and have `create_llm_backend` use the preset default when no explicit `timeout` is passed (explicit timeouts still override).

## Tests (`test_extraction_failure_policy.py`)

- extraction failure raises by default; degrades only with `extraction_on_failure="degrade"` (then `ok is False`); invalid policy rejected.
- `kimi` backend gets >120s timeout; `openai` keeps 120s; explicit timeout overrides the preset.

Both representative tests fail on `main`. `test_indexing.py` (59) stays green; the unrelated `graph.py` `ImportError` failures pre-exist on `main`.

Closes #118.


tteon:

PR #280, opened 2026-06-11
fix(store): Ladybug cross-document entity upsert — MERGE on declared PK (seocho-8ct)

tteon:
## Feature

Cross-document entity upsert for `LadybugGraphStore.write()`: MERGE now keys
on the table's declared PRIMARY KEY column instead of always `{id:}`, and the
merge key is excluded from the property-update map.

## Why

`ensure_constraints` declares the first unique ontology property (usually
`name`) as the Ladybug table PRIMARY KEY, while `write()` MERGEd on `id`.
Ladybug additionally rejects `SET` on a PK column, so every MERGE raised and
silently fell back to CREATE. When two documents mentioned the same entity
under different LLM-generated ids (`company_1` vs `company1`), that CREATE
violated the name PK and the whole file was reported failed
(`seocho run examples/run/quickstart.yaml` — sector_notes.md, "Found
duplicated primary key value Acme Corp"). Found during seocho-0u7 live e2e
verification; ticket seocho-8ct.

## Design

- `_node_table_pk` records the PK per node table in both creation paths
  (`ensure_constraints`, `_ensure_node_table`); tables discovered from an
  existing DB file default to `id`.
- `write()` MERGEs on the PK column when the node carries it, else `id`.
- The merge key is excluded from `SET n = {...}` (dialect probe confirmed
  Ladybug errors with "Cannot set property ... used as primary key"); the
  MERGE pattern pins it.
- The CREATE fallback keeps the full property map — creation must carry the PK.

## Expected Effect

Two documents mentioning the same entity upsert into one row; cross-document
relationships bind; per-file ingest no longer fails on the name-PK collision.

## Impact Results

- New repro test passes: doc1 (`company_1`) + doc2 (`company1`) for
  "Acme Corp" → exactly one Company row, `WORKS_AT` relationship created,
  zero errors.
- MERGE now actually merges (previously every MERGE fell back to CREATE).

## Validation

- `pytest tests/seocho/test_ladybug_store.py` → 17 passed, 2 failed — the 2
  failures are the pre-existing real_ladybug 0.15.3 parameter-expression
  regression tracked as seocho-g85, unrelated to this change (same 2 fail on
  main).
- `pytest tests/seocho/test_graph_ensure_database.py
  tests/seocho/test_graph_writer_lww.py tests/seocho/test_extraction_engine.py
  tests/seocho/test_e2e_runner.py tests/seocho/test_run_spec.py` →
  39 passed, 1 skipped.
- `py_compile src/seocho/store/graph.py` OK.
- Skipped: full `run_basic_ci.sh` locally (subset above covers the touched
  surface); ruff (not in venv).

## Risks

Low-medium. MERGE semantics change from "always CREATE via fallback" to true
upsert — node counts in `nodes_created` now count matched-and-updated rows
too (same as before, both paths incremented). Same-name nodes across labels
still resolve per-label tables. Pre-existing DB files (PK unknown) keep the
old id-keyed behavior.

### DIFF
--- a/src/seocho/store/graph.py
+++ b/src/seocho/store/graph.py
@@ -1354,6 +1354,12 @@ class LadybugGraphStore(GraphStore):
         self._declared_rel_tables: set = set()
         self._semantic_rel_types: set = set()
         self._rel_signature_to_table: Dict[tuple[str, str, str], str] = {}
+        # seocho-8ct: PRIMARY KEY column per node table. MERGE must key on
+        # the declared PK — Ladybug rejects MERGE on a non-key column, and
+        # the CREATE fallback then collides when two documents mention the
+        # same entity under different LLM-generated ids. Tables discovered
+        # from an existing database file fall back to "id".
+        self._node_table_pk: Dict[str, str] = {}
         self._load_existing_schema()
 
     def _locked_execute(self, *args, **kwargs):
@@ -1404,6 +1410,7 @@ class LadybugGraphStore(GraphStore):
                 f"CREATE NODE TABLE IF NOT EXISTS `{label}` ({col_list}, PRIMARY KEY (`{pk}`))"
             )
             self._declared_node_tables.add(label)
+            self._node_table_pk[label] = pk
         except Exception as exc:
             logger.warning("Failed to create node table %s: %s", label, exc)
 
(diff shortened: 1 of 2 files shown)


tteon:
Superseded by #282 — a parallel implementation of seocho-8ct landed on main first (same root cause: MERGE keyed on the declared PK; 21 ladybug tests green on main).

PR #284, opened 2026-06-11
feat(index,ontology): closed-vocabulary enforcement — strict/guided/open EnforcementPolicy (seocho-snt)

tteon:
## Feature

Closed-vocabulary ontology enforcement semantics: a frozen `EnforcementPolicy`
dataclass with `strict`/`guided`/`open` presets drives every enforcement
decision in the indexing path — prompt shaping, relaxed-retry gating,
heuristic-fallback gating, closed validation, out-of-ontology annotation, and
post-link re-validation.

## Why

Follow-up to the YAML e2e runner (seocho-0u7, PR #279): v1 mapped
`ontology.enforcement` onto the single `strict_validation` boolean, which
cannot express the panel-designed semantics — no Entity exemption, dangling
endpoints, domain/range conformance, no out-of-vocabulary fallback, and an
annotate-don't-reject open mode. Ticket seocho-snt.

## Design

- `src/seocho/index/enforcement.py`: `EnforcementPolicy` + `STRICT`/`GUIDED`/
  `OPEN` presets + `resolve_enforcement` (explicit policy or preset name wins,
  legacy `strict_validation` maps to strict/guided, default guided).
- `Ontology.validate_extraction(closed=)`: closed mode removes the `Entity`
  exemption, flags dangling endpoints via `known_ids`, and checks
  domain/range conformance through the `broader` chain (cycle-safe walk).
  `validate_with_shacl` forwards `closed`. Default behavior unchanged.
- `CanonicalExtractionEngine`: strict appends `_CLOSED_VOCABULARY_PROMPT_LINE`
  — constant and ontology-independent, so the extraction firewall
  (`to_extraction_context` byte-identity, FinDER r=-0.76 driver) holds and the
  KV-cache prefix stays stable per mode. Strict also disables the relaxed
  retry, which by design invites near-miss labels and generic Entity nodes.
- `IndexingPipeline`: strict gates `_fallback_extract` at both call sites
  (LLM failure and empty extraction) — the chunk is rejected with a recorded
  reason, never rescued with out-of-vocabulary structure. Validation runs
  `closed` under strict, and re-runs after the link step (linking can rewrite
  labels/endpoints). Rejection unit is the chunk — no silent element drops.
  Open mode annotates unknown labels/types with `_out_of_ontology: true`.
- Back-compat: `pipeline.strict_validation` is now a property over the
  policy; the setter flips strict/guided presets and no-ops when the boolean
  view already matches, preserving custom policies through the
  save/flip/restore pattern used by `local_engine`.
- Run-level failure thresholds intentionally remain in the e2e runner.

## Expected Effect

`enforcement: strict` runs produce only in-vocabulary graph structure (or
explicit chunk rejections); `open` runs keep exploratory structure but make
it filterable. `guided` is byte-for-byte today's behavior.

## Impact Results

- 15 new tests covering policy resolution, closed validation (Entity
  exemption, dangling endpoints, broader-chain domain/range, cycle safety),
  strict prompt-line + retry gating, pipeline fallback gating, open
  annotation, and the strict_validation property round-trip.
- Firewall regression suite untouched and green.

## Validation

- `bash scripts/ci/run_basic_ci.sh` → 569 passed, 1 skipped (new tests
  registered in the gate).
- `pytest tests/seocho/test_ontology_extraction_firewall.py
  tests/seocho/test_extraction_engine.py tests/seocho/test_e2e_runner.py
  tests/seocho/test_run_spec.py tests/seocho/test_ontology.py` → 77 passed.
- `git diff --check` clean.
- Skipped: ruff (not in venv); live LLM run (semantics covered by fakes).

## Risks

Medium. The indexing hot path changes for strict mode only; guided (the
default) resolves to a policy whose every gate matches current behavior, and
the full CI suite confirms no drift. The strict_validation setter replaces a
custom policy with a preset when actually toggled — documented in the
docstring, and the no-op guard covers the only in-repo flip pattern.

### DIFF
--- a/scripts/ci/run_basic_ci.sh
+++ b/scripts/ci/run_basic_ci.sh
@@ -50,6 +50,7 @@ python3 -m py_compile \
   src/seocho/tracing.py \
   src/seocho/store/graph.py \
   src/seocho/query/cypher_builder.py \
+  src/seocho/index/enforcement.py \
   src/seocho/index/extraction_engine.py \
   src/seocho/index/file_reader.py \
   src/seocho/run_spec.py \
@@ -117,6 +118,7 @@ uv run pytest \
   tests/seocho/test_ontology_iso704_cq.py \
   tests/seocho/test_run_spec.py \
   tests/seocho/test_e2e_runner.py \
+  tests/seocho/test_enforcement.py \
   -q
 
 git diff --check
--- a/src/seocho/index/__init__.py
+++ b/src/seocho/index/__init__.py
@@ -16,6 +16,7 @@ from .pipeline import (
     content_hash,
 )
 from .chunk import Chunk, build_chunk_id, chunk
+from .enforcement import GUIDED, OPEN, STRICT, EnforcementPolicy, resolve_enforcement
 from .extraction_engine import CanonicalExtractionEngine
 from .ingestion_facade import IngestRequest, IngestionFacade
 from .file_reader import (
@@ -29,6 +30,11 @@ from .file_reader import (
 __all__ = [
     "IndexingPipeline",
     "CanonicalExtractionEngine",
+    "EnforcementPolicy",
+    "resolve_enforcement",
+    "STRICT",
+    "GUIDED",
+    "OPEN",
     "IngestRequest",
     "IngestionFacade",
     "IndexingResult",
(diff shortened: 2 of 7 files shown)


tteon:
Superseded by #283 — a parallel implementation of seocho-snt landed on main first with equivalent-or-better coverage (incl. Any wildcard in domain/range conformance, post-link revalidation, dedicated tests). Closing to avoid a duplicate surface.

PR #408, opened 2026-07-13
Optimize file IO and JSON overhead in benchmarks and scripts

tteon:
### What
Replaced widespread usage of `path.read_text().splitlines()` with streaming file iteration (`with open(...) as f: for line in f`) across benchmark, CI, and tooling scripts.

### Why
`path.read_text().splitlines()` causes an $O(N)$ memory spike by buffering the entire file content into a single string in memory, and then splitting it into an array of strings, all before parsing the individual JSON lines. For large JSONL datasets or log files, this creates significant unnecessary memory overhead and slows down file parsing operations. Streaming file lines one-by-one is vastly more memory-efficient ($O(1)$) and avoids redundant string allocations.

### Expected Impact
Greatly reduced memory consumption and startup overhead for benchmarking and profiling scripts parsing large JSONL/log datasets. Also explicitly added `encoding="utf-8"` in places previously relying on system defaults to prevent platform-specific `UnicodeDecodeError` bugs on environments like Windows.

### Validation
Verified changes by running the local CI script `bash scripts/ci/run_basic_ci.sh`, which exercises the core logic and scripts. No regressions in functionality.

### Risks
Extremely low risk. Behaviorally equivalent to previous logic because `json.loads()` inherently ignores the trailing newlines preserved by file line iterators, meaning `line.strip()` isn't strictly required for JSON parsing, and any skipped blank lines still safely bypass decoding logic.

---
*PR created automatically by Jules for task [2789798652181342873](https://jules.google.com/task/2789798652181342873) started by @tteon*


tteon:
Closing: the perf content already landed in #419 (`fac6f941`), and this branch carries 366 files / +36,255 lines of unrelated merge history off a stale base.

PR #425, opened 2026-07-23
Optimize memory usage for file loading

tteon:
**What**
Replaced `Path.read_text().splitlines()` with a lazy file iterator (`with path.open() as f: for raw in f:`) in several maintenance scripts (`scripts/beads-path-guard.sh`, `scripts/gt-doctor.sh`, `scripts/task-context-trail.sh`) that use embedded Python code.

**Why**
Calling `read_text().splitlines()` loads the entire file into memory as a single string and then duplicates it into a list of strings. This creates avoidable JSONL and file I/O overhead. Iterating line-by-line prevents large files from causing memory spikes, specifically improving the `task-context-trail.sh` which often reads large JSONL events.

**Expected Impact**
Qualitative only: The change strictly preserves existing functionality but makes the script considerably more memory-efficient when handling very large config or event log files by streaming them lazily.

**Validation**
- `bash scripts/ci/run_basic_ci.sh` passed successfully.
- Code review validated that the logic exactly matches original loop behavior with preserved string formatting semantics, after a manual indentation fix.

**Risks**
Very low. Standard `path.open()` matches `path.read_text()` default file opening logic, and line processing behavior is kept identical via explicit line ending trimming (`raw.rstrip("\n")`).

---
*PR created automatically by Jules for task [15799096307716180170](https://jules.google.com/task/15799096307716180170) started by @tteon*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,1 +1,2 @@
 - Optimized memory usage for large JSONL files by replacing Path.read_text().splitlines() with lazy, line-by-line iteration using a context manager (with open(...) as f: for line in f:).
+- When applying automated patch scripts that modify pure Python files (especially involving indentation or loop structures), explicitly verify the syntax using `python -m py_compile <file>` to catch `SyntaxError`s early. Do not run `py_compile` directly on shell scripts (`.sh`) that embed Python code, as the surrounding bash syntax will cause false syntax errors.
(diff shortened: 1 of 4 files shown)


tteon:
Closing: this change already landed in #454 (`5455b277`, Optimize file loading).

PR #444, opened 2026-07-30
[Draft] Optimize JSONL read

tteon:
### What
Replaced inline `Path.read_text().splitlines()` with lazy file iteration (`with path.open() as f: for line in f:`) when parsing JSONL files in `scripts/benchmarks/finder_full_corpus_profile.py`. 

### Why
`Path.read_text().splitlines()` reads the entire JSONL file into memory at once as a single string, and then allocates a massive list of strings. For large JSONL corpus files, this creates a severe and entirely avoidable memory overhead (O(N) peak memory footprint instead of O(1)). Using an explicit loop and context manager allows line-by-line processing, effectively addressing one of the core repository performance priorities ("avoidable file I/O or JSONL overhead").

### Expected Impact
Greatly reduced peak memory footprint and elimination of unnecessary large-scale memory allocations when processing large JSONL files during benchmarks/corpus profiling. This optimization is primarily targeted at mitigating OOM errors on large datasets (qualitative impact).

### Validation
- Validated syntax with `python -m py_compile scripts/benchmarks/finder_full_corpus_profile.py`.
- Successfully ran the core validation script (`uv pip install -e .[ci,local] && bash scripts/ci/run_basic_ci.sh`) passing all 661 tests.
- Confirmed the behavior remains identical because `open("r")` defaults to universal newlines, and stripping trailing newlines `rstrip('\n')` cleanly simulates the behavior of `splitlines()` for each line, and `json.loads` is resilient to remaining whitespace.

### Risks
Minimal. File handling translates cleanly to lazy streaming, and exceptions are gracefully handled within the inner loop exactly as before.

### Modified Files
- `scripts/benchmarks/finder_full_corpus_profile.py`

---
*PR created automatically by Jules for task [6659986581680369038](https://jules.google.com/task/6659986581680369038) started by @tteon*

### DIFF
--- a/scripts/benchmarks/finder_full_corpus_profile.py
+++ b/scripts/benchmarks/finder_full_corpus_profile.py
@@ -50,11 +50,13 @@ def open_extract_all(jsonl_path, *, model, workers, max_chars, resume):
     jp = Path(jsonl_path); jp.parent.mkdir(parents=True, exist_ok=True)
     done_ids = set()
     if resume and jp.exists():
-        for line in jp.read_text(encoding="utf-8").splitlines():
-            try:
-                done_ids.add(json.loads(line)["id"])
-            except Exception:
-                pass
+        with jp.open("r", encoding="utf-8") as f:
+            for line in f:
+                line = line.rstrip('\n')
+                try:
+                    done_ids.add(json.loads(line)["id"])
+                except Exception:
+                    pass
     docs = []
     for _, row in df.iterrows():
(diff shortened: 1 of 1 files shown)


tteon:
Closing: the only real target in this cluster (finder_full_corpus_profile.py, 5,703-record JSONL), but the same function opens with a parquet download of the full FinDER corpus and then makes 5,703 LLM calls. ~5 MB of transient peak against that is not worth a merge.

PR #465, opened 2026-08-08
chore(lint): enforce ruff check repo-wide (pyflakes) and fix all violations

tteon:
## What

Ruff was already half-adopted: `[tool.ruff]` existed with 4 handpicked F-rules and `run_basic_ci.sh` ran `ruff check` on a small path allowlist, with a comment saying "when ready to enforce, add a CI step". This PR finishes that adoption for the lint side (beads: seocho-crd):

- `[tool.ruff] lint.select` → full pyflakes (`F`) set. Pyflakes-only keeps the CI gate on real-error class rules (unused/undefined names, redefinitions); no formatting-style rules, so external contributor PRs are not gated on style debates.
- `run_basic_ci.sh` now runs `ruff check src/seocho extraction runtime scripts tests` repo-wide (Basic CI already invokes this script on every PR).
- All 191 violations fixed: 159 auto-fixed unused imports, 32 reviewed by hand.

## Real bugs the gate caught

- **`ExecutionResult.agent_pattern` returned a `property` object** (`src/seocho/models.py`): a property shadowed the same-named dataclass field; with `slots=True` the property object became the field default and the property itself was stripped, so it never ran. All four construction sites pass the kwarg explicitly, so the dead property is deleted and the field default (`{}`) is restored. Regression test added (`tests/seocho/test_execution_result_models.py`, registered in basic CI).
- **`runtime/agent_server.py` imported `get_databases_impl`/`get_graphs_impl`/`get_schema_impl` and immediately shadowed them with local defs** — the imports were never effective; dropped them (runtime behavior unchanged, local defs still win).
- **`tests/seocho/test_ontology.py` had a duplicate `TestValidation` class** that shadowed the real one and referenced undefined names — it never ran and could not run; removed. (Its `validate_with_shacl` intent has no coverage elsewhere — noted for follow-up.)

## Re-export guardrail

The autofix initially stripped `build_evidence_bundle`/`infer_question_intent` from `seocho/query/answering.py`, which five downstream modules import *through* answering. Restored as explicit redundant aliases (`import x as x`) so ruff treats them as intentional re-exports. A repo-wide scan found no other removed import that anything else consumed. `store/__init__.py` similarly gains the two public exceptions in `__all__`.

## Deliberately out of scope (ticketed in beads)

- `ruff format` adoption: 543/691 files would be reformatted — big-bang needs coordination with open branches/WIP.
- Removing flake8/black/isort deps + `[tool.flake8]` (nothing in CI runs them today).

## Validation

- `bash scripts/ci/run_basic_ci.sh` → exit 0 (ruff: All checks passed; pytest: 663 passed, 5 skipped; contract checks all passed)
- `git diff --check` clean (runs inside the script)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tteon:
Queue-review status: still wanted, but this branch is unrebasable in practice (133 files, 2+ weeks of heavy churn since; main counts 175 repo-wide violations today, 150 auto-fixable). Plan: redo on current main (ruff --fix + manual pass + flip the CI gate to repo-wide) as the LAST merge of a quiet window — landing a repo-wide lint sweep now would force conflicts onto every in-flight branch (#621, #497, external rebases). Keeping this open as the tracking PR until the redo lands.

tteon:
Backlog review: draft pending regeneration on current main. This 133-file lint patch is based on an older repository surface and conflicts with subsequent ownership/runtime/test changes. Keep the intended F-rule cleanup, but regenerate it against current main and separate mechanical unused-import removal from behavioral changes; compare the resulting diff and run the full current basic CI. The local experiment checkout also contains newer lint work, which has been preserved separately rather than overwritten or silently published. No broad rebase conflict resolution should restore old module contents.

PR #481, opened 2026-08-14
refactor(observability): one metrics pipeline — ADR-0144 counters join the ADR-0146 registry

tteon:
**Stacked on #479** — retarget to `main` after it merges. Closes **seocho-8i0**.

Two disconnected pipelines carried metrics: `tracing.record_metric` (private OTel meter, snake_case names, no label validation, its own env/endpoint) and `metrics.ProductionMetrics` (catalog, allow-list + PII deny-list, histogram views, `SEOCHO_METRICS_BACKEND`). Enabling one did not enable the other, and `record_metric` bypassed the label budget entirely.

The eight ADR-0144 counters move into the catalog under dotted names (`seocho.index.validation_errors.count`, `.observations_reified.count`, `seocho.arbiter.route.count`, `seocho.query.plan/.db_hits/.scan/.plan_route/.generation_declined.count`) and their call sites (index/pipeline, plan_quality, local_engine) emit through `get_metrics()` directly. The tracing module's private MeterProvider is deleted — the OTLP backend owns spans only, and one env switch governs all metric export.

`record_metric` stays as a shim translating the legacy names; uncataloged names are **dropped**, not emitted — an arbitrary name is exactly the label-budget bypass being closed. None of the legacy series were referenced by any dashboard or alert rule (verified in the audit, held by `test_observability_contract`).

ADR-0144 amended in place (§6 note) + DECISION_LOG entry.

Validation: `bash scripts/ci/run_basic_ci.sh` — 681 passed, 3 skipped; `check-doc-contracts.sh` green; ruff clean.


tteon:
Closed automatically when its base branch was deleted; every commit landed on main via #482 (fix/saturation-signals contained this branch's full history). No content was lost.

PR #497, opened 2026-08-15
feat(serve-track): attribute vLLM KV blocks to graph-agentic-RAG stages

tteon:
Instrument for the PyTorch **Serve Track** talk: take a graph-agentic-RAG run apart stage by stage and say which stage caused which KV-cache behaviour.

## Why a sidecar

Neither side can attribute a block to a stage alone:
- vLLM's KV events **name no request** — `BlockStored` is `block_hashes + parent_block_hash + block_size + medium`.
- Trace schema v1 has `latency_ms` per step but **no absolute start time**, and it was frozen in #486.

So `kv_windows.py` writes a sidecar with the one fact both lack — when each step ran, on the same `time.time` clock the KV subscriber stamps frames with — and `correlate_kv.py` joins by containment. Sound only for serialized calls, which `WindowRecorder` enforces by refusing to open a second window while one is open.

## Three wrong assumptions the live runs caught

1. **vLLM 0.27.1 does not populate `prompt_tokens_details.cached_tokens` at all.** The first smoke asserted on it and failed while reuse was plainly happening. The usable signal is block-side — and better: `BlockStored` carries `block_size` (16), so blocks convert to tokens exactly. `cached_tokens` stays as the MARA signal, where there are no block events.
2. **Zero blocks stored = 100% reuse, the headline result.** The first `None` guard was per stage and reported it as "unmeasured". The guard belongs at run level.
3. **The smoke was not idempotent.** The previous run's blocks were still resident, so the cold call reused them and it read as "no reuse". The prefix is now salted per invocation.

## Measured — 1x RTX 3070 / Qwen3-0.6B, zero events unattributed

| stage | blocks | tokens stored | prompt tokens | reuse | latency |
|---|---:|---:|---:|---:|---:|
| `retrieve_ctx` (cold) | 55 | 880 | 885 | 0.6% | 789 ms |
| `synthesize` (warm) | 1 | 16 | 884 | **98.2%** | 542 ms |

## Portable by construction

`launch_vllm.sh` takes model, TP size, and memory fraction from env, so the **4x H200 / MiniMax-M2.7** measurement run uses identical code:

```bash
SERVE_MODEL=MiniMaxAI/MiniMax-M2.7 SERVE_TP=4 SERVE_GPU_MEM=0.90 scripts/serve_track/launch_vllm.sh
```

**KV offloading is already observable**: `BlockStored`'s `medium` field is reported as a per-stage `media` breakdown, so blocks recalled from CPU/disk show up without a new instrument.

## Scope

Does **not** claim the two caches can be co-managed. ADR-0156 recorded H0's verdict that the Neo4j page cache and the KV cache hold diverging working sets; the runbook says so explicitly.

## The real pipeline is now wired

`seocho.observability` gained the seam: `StageTimer.stage` publishes the active stage on a contextvar, and `OpenAICompatibleBackend.complete` (sync + async) wraps its call in `observe_llm_call` — a no-op until `set_llm_call_observer` is called. Opt-in is a function call, never an env var or import side effect; observer failures cannot fail a query, and a failed query still closes its window.

Stage routing goes through `StageTimer`, not the tracing span stack, because `start_span` returns a no-op and never pushes when no backend is configured, and its stack holds `(trace_id, span_id)` rather than names. `task_hint` was no good either — it is `"json_extraction"` for both text2cypher and entity extraction.

`profile_rag.py` drives `Seocho.ask()` against **DozerDB over Bolt** (the default — `Seocho.local()` routes to Neo4j/DozerDB only for a Bolt URI; LadybugDB ships a Bolt server but SEOCHO uses it as an in-process file store, so it is the wrong rig for a serving path).

### Measured on the real pipeline — and the zero is the finding

DozerDB + vLLM 0.27.1 / Qwen3-0.6B, 3 questions x 2 passes, 12 windows:

| stage | calls | engine prefix hit | `stable_prefix_chars` | latency |
|---|---:|---:|---:|---:|
| `plan` (text2cypher) | 6 | 55.5% | **0** | 264 ms |
| `generation` (synthesize) | 6 | 76.9% | 639 | 608 ms |

`plan` carries the ~3.2 KB ontology schema pack against a ~55-char question, yet its system section is a **different length for every question** (3238 / 3239 / 3288). Nothing is byte-stable across questions, so its hits come only from repeating the same question — per-question the warm pass reaches ~99%, but cross-question cold reuse is 8–14%. `generation`'s system section is identical across all six calls and reuses accordingly.

The engineering target is therefore upstream of the engine: make the schema pack byte-stable and put it first.

## Per-stage vLLM metrics

`vllm_metrics.py` samples `/metrics` at each window boundary and differences it, turning process-wide counters into per-stage numbers: offload bytes and transfer time, prefix-cache hits/queries, KV occupancy. Names were read off the installed 0.27.1 rather than the docs — notably **`vllm:kv_offload_total_bytes` is deprecated there**, superseded by the `store_bytes`/`load_bytes` + `store_time`/`load_time` split, which is the more useful shape (offloading pays off only when recall outweighs spill).

## Validation

- `python3 -m pytest tests/seocho/test_serve_track_correlation.py` — 8 passed
- `scripts/serve_track/smoke_correlation.py` against live vLLM 0.27.1 — passes
- `python3 -m ruff check scripts/serve_track/ tests/...` — clean
- vLLM stopped after the runs; GPU back to 75 MiB

Env note: vLLM 0.27.1 needs **Python 3.12**; on 3.10 flashinfer's `fd_exchange.py` subscripts `array.array` and fails at import.

---

## Follow-up: per-request plugin, structured decoding, real graph

**A stat-logger plugin replaces the containment join.** `FinishedRequestStats` carries `request_id`, `num_cached_tokens`, `num_prompt_tokens` and a queue/prefill/decode breakdown, so `scripts/serve_track/vllm_plugin` registers under the documented `vllm.stat_logger_plugins` group rather than patching `vllm.v1.core.kv_cache_coordinator` — which would reach nothing anyway, since `EngineCoreProc` runs in its own process. Verified live:

| request | prompt tokens | cached | prefill |
|---|---:|---:|---:|
| cold | 847 | 0 | 47.5 ms |
| warm | 847 | 832 (98.2%) | 11.7 ms |

That prefill pair is what an offloading decision turns on: recall pays only when moving the bytes beats the prefill it avoids.

**Structured decoding does not cost prefix reuse.** Plain, `response_format=json_object`, and two distinct `guided_json` schemas all reported **cached 832/847 (98.2%)**. The cost is one-time init — first guided request 325 ms vs ~85 ms plain, then a brand-new schema 96 ms vs 83 ms, a repeat indistinguishable. One request per arm, so treat the ms deltas as indicative.

**A correction.** My earlier note that MARA had extracted entities was wrong — it had been returning **401** the whole time. The default `guided` enforcement masked it: a failed extraction silently falls back to generic `Entity` nodes, so every `:Account` query matched nothing and synthesis got an empty record set while the run still reported success. `--enforcement` now defaults to strict, and `--index-llm` lets indexing run on MiniMax-M2.7 (what MARA serves, and what the H200 run will serve) while the measured serving path stays on `--llm`. The graph now carries `Account`/`Institution` and synthesis's user message grows 74 → 1681 chars on the question that retrieves.

**`generation` needed no reorder** — its 639-char instruction block already precedes the volatile records. Its low ceiling is inherent to per-question records, not a misordering.

---

## Why two demo questions retrieved nothing — and what it uncovered

Chasing this found a **P1 correctness bug that is not in this PR's scope** (filed as **seocho-k5n**).

**Q1 "Which accounts transferred money to B2?"** — retrieval was never empty; it was *inverted*. The `relationship_lookup` query matches undirected (`-[r:TRANSFER]-`) but labels the anchor as `source` and the neighbour as `target` unconditionally, discarding `startNode(r)`/`endNode(r)`:

| reported_source | reported_target | actual edge |
|---|---|---|
| B2 | A1 | **A1 → B2** |
| B2 | C3 | B2 → C3 |

The model then answered *"no accounts transferred money to B2; B2 transferred to A1 and C3"* — the opposite of the graph, stated confidently. `anchor_role='target'` was inferred **correctly** and is referenced in the builder, but does not constrain direction on this path. Any same-label relationship (payments, ownership, AML transfers) hits this silently.

**Q3 "Which accounts form a transfer cycle?"** — a different cause. Every deterministic intent is anchored on a named entity (`entity_lookup`, `relationship_lookup`, `neighbors`, `path`, `count`, `list_all`, `financial_metric_*`); none expresses a structural pattern like a cycle. The planner is forced to invent an anchor and picked the literal string `"Account"` — the label name — which matches no node. Not a bug so much as a missing intent class.

**Also confirmed:** the demo ontology declared no `source_role`/`target_role`, so `role_block` was empty and the prompt never even asked for `anchor_role`. Declaring `sender`/`receiver` made `anchor_role='target'` come out reliably (2/2 runs). Changing the ontology invalidates the indexed context hash, so a re-index is required to test it — the guardrail working as designed.

**Correction to my own diagnosis:** I first read `last_query_metadata['records']` as 0 and concluded retrieval had failed. It had not — the answer text showed the rows plainly. The metadata reading was the unreliable part, not the retrieval.

---

## The observability that was switched off

Tracing was **off for every measurement in this branch** — no `SEOCHO_TRACE_BACKEND`, and the harness never set one, so `start_span` returned a no-op throughout.

That is what produced the wrong diagnosis above. `rag.retrieve_ctx` records `n_records` directly; with a backend configured the question answers itself in one line:

```
rag.retrieve_ctx   {'n_records': 2, 'intent': 'relationship_lookup'}
rag.synthesize     {'result_count': 2}
```

Two records, not zero. Re-running the same set with tracing on also shows **Q3 retrieving 3 records** via `list_all` — so "retrieves nothing" was wrong there too, drawn from one non-deterministic run. Q3's real problem is narrower than stated: it retrieves, but `list_all` of accounts does not answer "which form a cycle", because no deterministic intent expresses a structural pattern.

`profile_rag.py` now defaults to `--trace-backend jsonl` and writes `spans.jsonl` into every run directory. The KV rig and the span tree measure different things and both are cheap.

**Honest limit:** the span tree counts records, it does not carry their content — so it could not have revealed the inverted edge direction (seocho-k5n). Counts catch "did anything come back"; correctness needs the values.


tteon:
Queue-review status: deferred, not forgotten — this is the PyTorch Serve Track instrumentation (54 files, +8.5k) and partially overlaps the harness that landed via #561. It needs its own review cycle with the talk timeline in mind rather than a queue sweep; also worth checking which pieces #561 already covers before rebasing.

tteon:
Backlog review: keeping only the unlanded SDK/scorecard proposal as draft. The harness was already landed in #561 with the ERB answer-key leakage fixed. That review explicitly excluded ontology_modularity/workflow_scorecard/plan-quality additions: the modularity metrics rated a flat ontology optimal and the scorecard lacked a production caller. Main retains the corrected extracted-evidence default; do not merge this older branch over it. Rebuild a narrow PR for the remaining SDK behavior, address those findings, and link traceable live evidence. Existing reports from the old answer_facts arm must not be cited as retrieval or quality wins.

PR #160, opened 2026-05-30
fix(examples/finder): T4 notebook fails on seocho 0.4.1 — from_ttl() has no name= kwarg

Shin-E2:
## Problem

`examples/finder/04_private_opik.ipynb` (cell 5) calls:

```python
base = Ontology.from_ttl(TTL_DIR / "private_base.ttl", name="private_base")
```

but `seocho.ontology.Ontology.from_ttl(path)` (seocho 0.4.1, the version the tutorial Dockerfile installs / the repo source at this commit) takes no `name` argument:

```
TypeError: from_ttl() got an unexpected keyword argument 'name'
```

So `make tutorials-test` fails on Tutorial 4 (T1/T2/T3 pass) — the shipped tutorial cannot run against the library it ships with.

## Fix

Set `.name` after loading. The same cell already does `composed.name = ...` a few lines later, so `.name` is a writable attribute:

```python
base       = Ontology.from_ttl(TTL_DIR / "private_base.ttl"); base.name = "private_base"
extension  = Ontology.from_ttl(TTL_DIR / "private_extension.ttl"); extension.name = "private_extension"
restricted = Ontology.from_ttl(TTL_DIR / "private_restricted.ttl"); restricted.name = "private_restricted"
```

## Verification

- `jupyter nbconvert --to notebook --execute examples/finder/04_private_opik.ipynb` now completes with 0 error cells.
- Ontology `+` / `-` compose still works (Composed labels: FinancialMetric, LegalEntity, Patent, Person, Regulator).
- Opik traces upload as before.

Tested against seocho 0.4.1.

Found while reproducing the FinDER tutorials for the OSSCA 2026 GraphRAG mentorship — thanks for the great teaching material!

PR #508, opened 2026-08-15
fix: five legacy crash fixes rescued from the stale fork (#234 #236 #239 #245 #160)

tteon:
Five independent crash/`TypeError` fixes from the `piso7/seocho` fork, where they sat unmergeable — branched off `590f8eb5` (2025-12-06), reported by GitHub as 900+ conflicting files each. Every one is a single clean commit; cherry-picked onto current `main` with authorship preserved, all five applied without conflict.

Each was verified as **still present on `main`** before rescuing.

| PR | Defect on main | Where |
|---|---|---|
| **#239** | `csv.DictReader(f)` with no `restval` → ragged rows give `content = None` → `content.strip()` raises `AttributeError` in the `index()` loop, which is not wrapped | `index/file_reader.py` |
| **#236** | `ask_stream` does `asyncio.new_event_loop()` + `run_until_complete` → `RuntimeError` inside Jupyter or a FastAPI handler. The sync `ask` path already has the correct `_run_sync` worker-thread pattern | `session.py` |
| **#245** | `IndexFlatIP(1536)` hard default with `self._dimension` stored and never checked → `text-embedding-3-large` (3072) or any vLLM/Qwen preset yields an opaque assertion from faiss's C layer | `store/vector.py` |
| **#234** | `input=list(texts)` unchunked in **both** embed implementations → a large `add_batch` 400s and drops the whole batch | `store/llm.py` |
| **#160** | `from_ttl(cls, path)` has no `name=` parameter; the T4 notebook passes one, so the cell raises `TypeError` today | `examples/finder/04_private_opik.ipynb` |

Four of the five carry a regression test from the original author. #160 is a three-line notebook fix.

## Validation

- all five cherry-picks clean against current `main`
- `bash scripts/ci/run_basic_ci.sh` — **768 passed**, 3 skipped
- the final commit registers `test_stream_async_in_thread.py` and `test_faiss_vector_store.py` in `run_basic_ci.sh`, which takes an explicit file list rather than a directory — otherwise the new guards would sit in the tree without ever executing

## Context

These come out of a triage of ~34 legacy fork PRs. Of those, **14 are live silent-wrong-result or silent-data-loss bugs**; this PR takes the five crash-class ones because they are independent and low-risk. The higher-stakes ones — dedup scoping (#243), inverted financial deltas (#240), LanceDB score polarity (#235), FAISS delete under-return (#246), stale vectors on `delete_source` (#261) — are next, and want more care.

Closes #234. Closes #236. Closes #239. Closes #245. Closes #160.


tteon:
Superseded — all five fixes already landed on main, four verbatim (same code, comments, and regression tests; the landed commits appear to be cherry-picks of this cluster): #239 via PR #548, #236 via #552, #245 via #549, #234 via #553; the #160 notebook fix is obsolete (the notebook was deleted with the Opik removal, PR #518). Decisive check: auto-merging this branch onto main leaves an EMPTY src/tests diff. The corresponding suites are green on main (62 passed).

PR #509, opened 2026-08-15
fix: five legacy silent-wrong-result bugs (#235 #240 #243 #246 #261)

tteon:
Five more single-commit fixes rescued from the `piso7/seocho` fork. Unlike #508 (crashes), **every one of these fails silently** — no exception, no log line, just a wrong answer or missing data. Each was verified as still present on `main` before rescuing; all five cherry-picked without conflict.

| PR | What goes wrong, silently | Where |
|---|---|---|
| **#243** | `_seen_hashes` is **instance** state, so a genuinely distinct document whose normalized text collides with an earlier one returns `deduplicated=True` and **writes nothing**. Behaviour depends on process lifetime — the same two documents index differently depending on whether they arrived in one run or two. | `index/pipeline.py` |
| **#240** | `financial_metric_delta` takes its endpoints from `_ordered_years`, which always sorts chronologically. *"How did revenue change from 2023 to 2021"* is answered as 2021→2023 — **sign inverted, increased/decreased flipped**. | `query/answering.py` |
| **#235** | `VectorSearchResult.score` is FAISS inner-product (higher is better) but LanceDB `_distance` (lower is better) — the same field means **opposite things** by backend. Any `sort(reverse=True)` consumer ranks LanceDB results **worst-first**. | `store/vector.py` |
| **#246** | `k = min(limit, ntotal)`, then deleted docs are filtered **after** the search, so `search(limit=k)` returns fewer than `k` live results even when more exist. Amplified because `add()` tombstones on re-add. | `store/vector.py` |
| **#261** | `delete_source()` only calls `graph_store.delete_by_source(...)`; `reindex()` goes through it. Old vectors survive and keep surfacing as top-k hits that resolve to deleted or stale graph nodes. | `index/pipeline.py`, `store/vector.py` |

## On #240 specifically

The fix does not simply reverse the sort. It follows the order the years were **stated in the question**, and falls back to chronological order only when the question named no years:

```python
phrased_years = self._dedupe_years(years)
if len(phrased_years) >= 2:
    start_year, end_year = phrased_years[0], phrased_years[-1]
else:
    target_years = self._ordered_years(years or [row["year"] for row in selected_rows])
```

## On #235 specifically

Distance is mapped to similarity with `1.0 / (1.0 + raw_distance)`, so both backends now put "better" at the high end and the field is comparable across stores. Note this is a **monotonic re-mapping, not a calibration** — absolute scores still are not comparable between FAISS and LanceDB, only their ordering is.

## Validation

- five cherry-picks clean against current `main`
- `bash scripts/ci/run_basic_ci.sh` — **771 passed**, 3 skipped
- final commit registers `test_faiss_delete.py`, `test_dedup_scope.py` and `test_financial_delta_direction.py` in `run_basic_ci.sh`, which takes an explicit file list rather than a directory

## Review note

These change behaviour, not just error paths, so they deserve more scrutiny than #508. In particular: #243 changes when a second identical-text document is written, and #261 adds a vector deletion that did not previously happen. If anything downstream relied on the old dedup being process-wide, this will surface there.

Closes #235. Closes #240. Closes #243. Closes #246. Closes #261.

### DIFF
--- a/scripts/ci/run_basic_ci.sh
+++ b/scripts/ci/run_basic_ci.sh
@@ -102,6 +102,9 @@ uv run pytest \
   tests/seocho/test_ontology_reasoner.py \
   tests/seocho/test_ontology_iso704_cq.py \
   tests/seocho/test_relationship_direction.py \
+  tests/seocho/test_faiss_delete.py \
+  tests/seocho/test_dedup_scope.py \
+  tests/seocho/test_financial_delta_direction.py \
   tests/seocho/test_prompt_prefix_stability.py \
   tests/seocho/test_run_spec.py \
   tests/seocho/test_e2e_runner.py \
--- a/src/seocho/index/pipeline.py
+++ b/src/seocho/index/pipeline.py
@@ -1429,6 +1429,19 @@ class IndexingPipeline:
 
         Returns
         -------
-        Summary with ``nodes_deleted``, ``relationships_deleted``.
+        Summary with ``nodes_deleted``, ``relationships_deleted`` (and
+        ``vectors_deleted`` when a vector store is attached).
         """
-        return self.graph_store.delete_by_source(source_id, database=database)
+        summary = self.graph_store.delete_by_source(source_id, database=database)
+        # Also drop the source's vectors so the vector store stays consistent
+        # with the graph; otherwise stale vectors keep surfacing as top-k hits
+        # pointing at deleted nodes (issue #123). reindex() goes through here.
+        if self.vector_store is not None:
+            try:
+                removed = self.vector_store.delete_by_source(source_id)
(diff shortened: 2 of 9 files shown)


tteon:
Superseded — all five fixes re-landed 2026-08-16 via #548/#549/#552, byte-identical or strictly better. One detail recorded so this branch isn't resurrected: the #261 LanceDB hunk here deletes with id LIKE '{src}_chunk_%' — in SQL LIKE, underscore is a single-char wildcard, so reindexing one document could silently destroy a neighboring source's vectors (e.g. report_chunk_% also matches report_chunked_v2_chunk_0000). Main's version uses an exact half-open range predicate, and its regression tests (test_delete_source_vectors.py) would fail against this branch's code.

PR #618, opened 2026-08-20
fix(query): _count anchors the filter entity on the ontology-declared side

jaewoong-newtonne:
## Symptom

On our production carbon-registry graph (17.5k `Project` nodes, `Project -LOCATED_IN-> HostCountry`), the question **"How many carbon crediting projects are registered in Vietnam?"** answered **0** (actual: 96), with `support_assessment.reason: graph_records_returned` — a plausible wrong answer rather than an error.

## Root cause

The executed query (captured at the bolt level) was:

```cypher
MATCH (src:`HostCountry`)-[:`LOCATED_IN`]->(anchor:`Project`)
WHERE (anchor.`registry_id` = $anchor_key_0)   -- $anchor_key_0 = 'Vietnam'
...
RETURN count(DISTINCT src) AS count
```

Two compounding effects in `_count`'s relationship-aware branch:

1. Plans routinely put the **counted** label in the anchor slot (`anchor_label=Project` for "how many projects…"), and `_orient_relationship` pins the anchor to the declared source — so the filter equality lands on the counted side (`Project.registry_id = 'Vietnam'`).
2. The default (non-`anchor_role=source`) pattern places the anchor at the arrow head, which combined with (1) emits the traversal **against** the declared direction.

Both plan shapes (`anchor=HostCountry` and `anchor=Project`) collapse into the same always-zero query. The failure is invisible on same-label schemas (FinBench `TRANSFER: Account -> Account`) where orientation repair is a no-op — it needs distinct endpoint labels to surface.

## Fix

When the relationship declares two **distinct** endpoint labels, the declaration is authoritative: counted side = source, anchored side = target (flipped under `anchor_role="source"`). Same-label relationships keep the existing role-based behaviour unchanged.

With the fix, both plan shapes emit:

```cypher
MATCH (src:`Project`)-[:`LOCATED_IN`]->(anchor:`HostCountry`)
WHERE (anchor.`country_iso` = $anchor_key_0 OR anchor.`name` = $anchor_key_1)
...
RETURN count(DISTINCT src) AS count
```

which returns 96 on our graph.

## Tests

- 2 regression tests added in `test_cypher_builder_ontology_aware.py` (distinct-label authority for both plan shapes; same-label in-degree behaviour preserved)
- `pytest tests/seocho -k "cypher or count or lpg or builder"`: 84 passed, 1 pre-existing failure (`test_workspace_filter_enforcement.py::test_enforce_workspace_filter_passes_when_cypher_references_param` — fails identically on 59d7028 without this change)

## Related observations (separate, happy to file issues)

- `_local_support_assessment` treats `row_count > 0` as `supported` — for count queries the single `{count: 0}` row made the zero answer look supported. A count-shape check would help.
- A single-token mention passes the `_sargable_anchor` identifier gate even when it is a display name ("Vietnam" vs `country_iso='VNM'`); we worked around it ontology-side with `index: true` on `HostCountry.name`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NnEzVBLjwxgEYFmDTrCaqw

### DIFF
--- a/src/seocho/query/cypher_builder.py
+++ b/src/seocho/query/cypher_builder.py
@@ -1039,11 +1039,31 @@ class CypherBuilder:
         # scan and report the total node count.
         if anchor_entity and relationship_type in self.ontology.relationships:
             src_label = target_label if target_label in self.ontology.nodes else label
-            src_clause = f":{quote_identifier(src_label)}" if src_label else ""
-            rel_clause = f":{quote_identifier(relationship_type)}"
             # Anchor label matters twice: it makes the pattern index-eligible and it
             # tells us which identity key to compare against.
             anchor_label = label if label in self.ontology.nodes else src_label
+            # When the ontology declares two DISTINCT endpoint labels, the
+            # declaration is authoritative for which side is counted and which
+            # side carries the anchor predicate. Plans routinely arrive with the
+            # counted label in the anchor slot ("how many projects in Vietnam"
+            # → anchor_label=Project), and _orient_relationship then pins the
+            # anchor to the declared source — so the filter equality landed on
+            # the counted side (anchor.registry_id = 'Vietnam') while the
+            # pattern ran against the declared direction reversed: a valid query
+            # that always matches zero rows. Same-label relationships (TRANSFER:
+            # Account -> Account) keep the existing role-based behaviour.
+            rel_def = self.ontology.relationships[relationship_type]
+            if (
+                rel_def.source in self.ontology.nodes
+                and rel_def.target in self.ontology.nodes
+                and rel_def.source != rel_def.target
+            ):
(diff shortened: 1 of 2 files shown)


tteon:
Backlog review: keeping this draft pending a bidirectional count fixture. The new tests pass, but the heuristic treats every distinct-label relationship without anchor_role="source" as a target anchor and overwrites the supplied anchor label. An ontology's edge direction does not prove which entity the question named. For Project -> HostCountry, both counting projects in Vietnam and countries associated with a named project must preserve the identity predicate on the correct entity. Cover both explicit roles, both correctly bound labels, missing roles, and inconsistent model-filled labels. Ambiguous inputs should be repaired upstream or rejected, not silently rebound to another entity type. Validate against a small live fixture before landing a quality claim.

PR #627, opened 2026-08-22
fix: Enforce workspace filter in Seocho.query()

tteon:
Severity: High

Vulnerability: Missing authorization checks on `Seocho.query()` in local mode. Direct graph store queries bypass the workspace filtering that restricts data access by tenant.

Impact: A user calling `query()` could inadvertently access or modify data belonging to other workspaces (tenant isolation bypass) if they pass malicious Cypher.

Fix: Added `workspace_id=self.workspace_id` and `enforce_workspace_filter=True` to the `self.graph_store.query()` call inside `Seocho.query()`.

Validation: Ran `bash scripts/ci/run_basic_ci.sh` to ensure no tests break and local tests verify the behavior.

Risks: Queries constructed without a workspace variable in the string may now correctly fail with a `WorkspaceFilterMissingError`. Users relying on global access (which is unintended in tenant environments) must opt out via `enforce_workspace_filter` or provide unscoped queries via lower-level driver directly.

Modified Files: src/seocho/client.py

---
*PR created automatically by Jules for task [1714042385666734580](https://jules.google.com/task/1714042385666734580) started by @tteon*

### DIFF
--- a/src/seocho/client.py
+++ b/src/seocho/client.py
@@ -1531,7 +1531,13 @@ class Seocho:
         """
         if not self._local_mode:
             raise RuntimeError("query() requires local engine mode (ontology + graph_store + llm)")
-        return self.graph_store.query(cypher, params=params, database=database)
+        return self.graph_store.query(
+            cypher,
+            params=params,
+            database=database,
+            workspace_id=self.workspace_id,
+            enforce_workspace_filter=True,
+        )
 
     # ------------------------------------------------------------------
     # Agent-level session API


tteon:
Consolidating workspace-enforcement review in draft #655. This proposal is not being represented as a merged security fix. The canonical validator, conflicting workspace parameters, scoped aliases and valid-query regressions need coordinated coverage; the detailed findings and acceptance gate are on #655. Closing the overlapping PR to keep one review thread.

PR #657, opened 2026-09-05
perf: use lazy file streaming for graphrag bench eval jsonl files

tteon:
This PR introduces a performance optimization to graphrag benchmarking evaluation.

**What**
Modified `src/seocho/eval/graphrag_bench.py` to stream JSONL evaluation files via lazy file iteration (`with path.open("r", encoding="utf-8") as f: for line in f:`) instead of eagerly reading the entire file into memory with `path.read_text(encoding="utf-8").splitlines()`.
Modified Files: `src/seocho/eval/graphrag_bench.py`, `.jules/bolt.md`

**Why**
Prior to this change, large evaluation datasets would be read entirely into memory as one monolithic string, then split into a large list of string objects before processing. This created avoidable and significant memory/GC overhead, especially during scaled evaluation runs.

**Expected Impact**
Lower memory usage peak during GraphRAG evaluation runs. Reduced latency before first row evaluation begins for very large benchmark sets.

**Validation**
Run `PYTHONPATH=. uv run pytest tests/seocho/test_benchmarking.py` locally and verified that the test suite passes, indicating that the behavior/parsing correctness was preserved. 
Ran `bash scripts/ci/run_basic_ci.sh` to ensure overall runtime sanity.

**Risks**
Minimal. `json.loads` safely ignores trailing newlines that are preserved by file iteration compared to `splitlines()`.

*(Draft PR as requested)*

---
*PR created automatically by Jules for task [1135595266862568347](https://jules.google.com/task/1135595266862568347) started by @tteon*


tteon:

PR #660, opened 2026-09-07
perf(eval): use lazy file iteration for JSONL loading

tteon:
**What:**
Replaced `path.read_text(encoding="utf-8").splitlines()` with `with path.open("r", encoding="utf-8") as f: for line in f:` in `load_question_directory` of `src/seocho/eval/graphrag_bench.py`.

**Why:**
`.jsonl` evaluation files can grow large. The previous approach read the entire file into memory as a single string, and then allocated a large list of strings for all lines via `splitlines()`, which causes unnecessary peak memory bloat.

**Expected Impact:**
Lower memory usage and faster processing when loading eval test sets, as the file is now lazily iterated over line-by-line.

**Validation:**
- Passed `tests/seocho/test_benchmarking.py` using `uv run pytest`.
- Logic strictly preserves existing behavior (as `json.loads` natively ignores trailing newlines that `splitlines` otherwise stripped).

**Risks:**
Extremely low risk as behavior is functionally identical.

**Modified Files:**
- `src/seocho/eval/graphrag_bench.py`
- `.jules/bolt.md`

*(Draft PR as requested)*

---
*PR created automatically by Jules for task [10741276729767273433](https://jules.google.com/task/10741276729767273433) started by @tteon*


tteon:

PR #667, opened 2026-09-10
perf: optimize jsonl trace loading with lazy streaming

tteon:
### What
Refactors `load_question_directory` in `src/seocho/eval/graphrag_bench.py` to stream JSONL evaluation trace files using a lazy `for line in f:` iterator rather than reading the entire file string and splitting it into a massive list via `path.read_text().splitlines()`.

### Why
During evaluation benchmarks (which use JSONL trace inputs heavily), loading the full file at once causes a drastic memory spike and avoidable allocation latency (O(N) vs O(1)). This addresses the priority area of "avoidable file I/O or JSONL overhead".

### Expected Impact
Qualitative: Significantly lower peak memory usage and marginally faster initialization times during benchmark dataset loads, specifically preventing out-of-memory overhead when processing large evaluation sets. Memory overhead during iteration changes from O(N) to O(1).

### Validation
Ran `pytest tests/seocho/test_eval_benchmark.py` and `bash scripts/ci/run_basic_ci.sh`. All tests pass successfully, confirming that the change introduces no regressions to trace processing.

### Risks
None. File locking/I/O behavior is handled safely by the `with` block context manager. `json.loads` safely ignores trailing newline elements correctly.

Modified Files: src/seocho/eval/graphrag_bench.py, .jules/bolt.md
Draft PR: Yes

---
*PR created automatically by Jules for task [9736674185303750288](https://jules.google.com/task/9736674185303750288) started by @tteon*


tteon:

